PR #4, opened 2014-01-08
Support any usage ids from the HID consumer page

gregersrygg:
The implementation in Teensy 3 (and previous) only supports 8 hard-coded consumer-keys, so it was not possible to send the key AC_HOME (0x223) or any of the other 660 consumer-keys. I need some of these consumer-keys to create a custom button-panel for my in-car Nexus 7. The home-button, back-button and many other Android functions can only be triggered by consumer-keys (reference http://source.android.com/devices/te...d-devices.html). Some of them are two bytes, so I found it was not sufficient to just modify the keyboard report descriptor.

I changed the keyboard report descriptor to use an array for the consumer-keys. Not sure if that's Correct™, but that's what I got working. Since some of the consumer page usage ids consists of 2 bytes, I also had to steal one byte from the 6 “normal” keyboard keys (now 5). It's also only possible to send one consumer key at the time. Not sure if there are enough available bytes in the packet to add more consumer-keys, or if it's better to add a new HID config for Consumer Keyboard?

Since set_media was only for media-related keys, I added a new method; set_consumer that can take a 16-bit int. Changed the media-key constants in keylayouts.h to the usage id, and set_media now just calls set_consumer.

Tested many different consumer keys on Android 4.2, and they work perfectly. Only tested the volume-keys on OS X 10.9.1, and they work as well. Tried to test on Windows in VirtualBox, but it wouldn't allow me to use the Teensy USB keyboard :( The AC_HOME even works as the home-button on an iPad if using the iPad camera connection kit!

### DIFF
--- a/teensy3/keylayouts.h
+++ b/teensy3/keylayouts.h
@@ -80,14 +80,14 @@ extern "C"{
 #define MODIFIERKEY_RIGHT_ALT   ( 0x40 | 0x8000 )
 #define MODIFIERKEY_RIGHT_GUI   ( 0x80 | 0x8000 )
 
-#define KEY_MEDIA_VOLUME_INC    0x01
-#define KEY_MEDIA_VOLUME_DEC    0x02
-#define KEY_MEDIA_MUTE          0x04
-#define KEY_MEDIA_PLAY_PAUSE    0x08
-#define KEY_MEDIA_NEXT_TRACK    0x10
-#define KEY_MEDIA_PREV_TRACK    0x20
-#define KEY_MEDIA_STOP          0x40
-#define KEY_MEDIA_EJECT         0x80
+#define KEY_MEDIA_VOLUME_INC    0xE9
+#define KEY_MEDIA_VOLUME_DEC    0xEA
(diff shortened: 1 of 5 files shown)


NicoHood:
Why are the consumer keys even in the descriptor if they dont work? Ive tested this once and it didnt work of course. (with a different avr, maybe that's the Problem??)

Is there any Option to get it working inside a Keyboard Report or does it have to be on a 2nd endpoint as you suggested?

PaulStoffregen:
TL;DR = When I wrote this, complying with the HID spec, I did not know Microsoft Windows requires a separate interface/endpoint.  It does fully comply with the spec and works great on Mac and Linux.  Only Windows doesn't work.  I will redo it someday....

NicoHood:
I suggest that you keep it like this. It maybe doesnt work with windows but i didnt know that it works with linux. The normally reseved byte can then be used with some usefull stuff there. And since its not in the api itself nobody get confused. I'd leave it like that if you ask me. Maybe a note would be nice so people know what this is used for.

PR #13, opened 2014-04-25
Fix print() for long long

FrankBoesing:
Printing huge integer numbers did not work; this patch repairs this.
Additionaly, in print() was uint8_t buf[34]; which was bigger than
nessesary.

Testsketch (did not work before):

#include <climits>

void setup(){
Serial1.begin(115200);
Serial1.println();
Serial1.println();
Serial1.println(LONG_MIN);
Serial1.println(LONG_MAX);
Serial1.println(ULONG_MAX);
Serial1.println(LLONG_MIN);
Serial1.println(LLONG_MAX);
Serial1.println(ULLONG_MAX);
}

void loop(){
}

### DIFF
--- a/teensy3/Print.cpp
+++ b/teensy3/Print.cpp
@@ -57,7 +57,7 @@ size_t Print::print(const String &s)
 }
 
 
-size_t Print::print(long n)
+size_t Print::print(long long n)
 {
 	uint8_t sign=0;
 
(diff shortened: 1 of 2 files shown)


PaulStoffregen:
I would like to support printing 64 bit integers, but not by enlarging and slowing the printNumber() function used for 32 bit and smaller.

This really needs to be done with another function like printNumber64.

PR #40, opened 2014-10-25
Use GCC atomic operations instead of IRQ disable.

notaz:
Avoids unnecessary interrupt latency caused by irq disable.
GCC atomics map to LDREX/STREX that M4 in teensy3 supports.

### DIFF
--- a/teensy3/usb_mem.c
+++ b/teensy3/usb_mem.c
@@ -47,22 +47,22 @@ static uint32_t usb_buffer_available = 0xFFFFFFFF;
 
 usb_packet_t * usb_malloc(void)
 {
-	unsigned int n, avail;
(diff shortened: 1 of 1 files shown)


FrankBoesing:
+1

Theremingenieur:
+1

jasa:
Not quite true,  download and read in the "ARM Architecture Reference Manual" before you make the same mistake as i did, there is a full page of restrictions.
<A2.9 Synchronization primitives>
Where a context switch or exception might result in a change of execution thread, a dummy STREX instruction, as described in Context switch support on page A2-48 should be executed
to avoid unwanted effects. This is the only occasion where an STREX is expected to be programmed
with a different address from the previously executed LDREX .

PR #50, opened 2015-02-04
Fix F_PLL

duff2013:
Shouldn't the F_PLL match the F_CPU when defined in kinetis.h? Anyway this fixes them if thats the case.


PaulStoffregen:
The PLL really is at 96 MHz in those cases.  The clock divider is used to reduce the PLL down to the CPU clock rate.

Until now, we never needed to know the PLL speed.  Everything ran from F_CPU or F_BUS.  But in Teensy-LC, a couple things run from PLL/2, regardless of the dividers for the CPU and BUS.

PR #51, opened 2015-02-10
(MKL26Z64) update for F_CPU <= 16MHz

duff2013:
Tested just Uart1 so far and there is a problem at 2 MHz with the baud rate divisor. I'm just not seeing how to produce the right divisor? Also haven't looked at any other peripherals so I'm sure they will need some help also. Here is list of what I did:

1. Updated Uart BAUD2DIV, problem @ 2 MHz with baud generation.
2. Fixed F_PLL value <= 16 MHz.
3. Enabled MCGIRCLK and OSRERCLK @ F_CPU <= 16 MHz.
4. Fixed divider, no need to set SIM_CLKDIV1_OUTDIV2 @ F_CPU <= 16 MHz.
5. Fixed SIM_SOPT2_CLKOUTSEL & SIM_SOPT2_UART0SRC clocks @ F_CPU <= 16 MHz.
6. Fixed F_TIMER, not sure this is right but had touch it to make it compile, might want to double check this.

### DIFF
--- a/teensy3/HardwareSerial.h
+++ b/teensy3/HardwareSerial.h
@@ -90,7 +90,15 @@
 #define BAUD2DIV2(baud) (((F_CPU * 2) + ((baud) >> 1)) / (baud))
 #define BAUD2DIV3(baud) (((F_BUS * 2) + ((baud) >> 1)) / (baud))
 #elif defined(KINETISL)
+
+#if F_CPU <= 2000000
+#define BAUD2DIV(baud)  (((F_PLL / 16 ) + ((baud) >> 1)) / (baud))
+#elif F_CPU <= 16000000
+#define BAUD2DIV(baud)  (((F_PLL / (F_PLL / 1000000)) + ((baud) >> 1)) / (baud))
+#else
 #define BAUD2DIV(baud)  (((F_PLL / 2 / 16) + ((baud) >> 1)) / (baud))
+#endif
+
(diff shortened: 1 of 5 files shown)


PaulStoffregen:
I suspect some of these SIM_CLKDIV1_OUTDIV4 settings are incorrect, but I'm going to merge this now, do a few quick tests to make sure 24 & 48 MHz still work, and then push out 1.21-beta6.

duff2013:
- I suspect some of these SIM_CLKDIV1_OUTDIV4 settings are incorrect

Is it not using 16MHz crystal? I tested sucessfully Serial1 at from 1200 - 115200 baud at 16,8,4,MHz so it would seem that those divisors should be correct. At 2 MHz only 9600 baud worked so I pulled it up on the scope others bauds would slightly dirft from what they should be. I used the serial decode on my scope and since I'm front of my scope for the next few days I can't tell you how off these bit streams where. Maybe the trim value for MCGIR clock needs to adjusted?

PaulStoffregen:
On the K series, both OUTDIV1 and OUTDIV4 divide down the PLL.

On L series chips, OUTDIV1 divides down the PLL.  OUTDIV4 gets its input from OUTDIV1.  So if you want the CPU and BUS+Flash clock the same speed, OUTDIV4 needs to be zero.

There's only a very small mention of this in the manual, and it's not worded very clearly.

If you configure it based on experience from the K20 chips, you end up with the processor running at the right speed, but peripherals and flash running very slowly.

PR #52, opened 2015-02-15
Migrate to Arduino 1.6.0

ThyMYthOS:
See also https://github.com/arduino/Arduino/wiki/Arduino-Hardware-Cores-migration-guide-from-1.0-to-1.6


PaulStoffregen:
Please use Teensyduino 1.21-beta5 for Arduino 1.6.0 support.

https://forum.pjrc.com/threads/27740-Arduino-1-6-0-any-plans-to-support-it

PaulStoffregen:
I made many design decisions differently than this pull request does.  The approach I used is somewhat different, with only a single boards.txt and platform.txt.  I do NOT wish to create 2 platforms for Teensy.  The reasons are explained on that forum thread.

ThyMYthOS:
I've read your posts, but I strongly disagree with ignoring the new structure to have one dir for each architecture. This undermines all the good approaches Arduino is now taking. How should the devs learn the new structure if the HW devs ignore it?

Please rethink your decision.

PaulStoffregen:
My main goal, more important that everything else, it to make as many sketches and libraries "just work" as possible.  Today, many people design libraries only for "avr".  I will not use any other directory name, as long as the Arduino world is filled with design only for avr architecture.

This issue was discussed at length on Arduino's developers mail list, about 1 year ago.  They did make several changes to their system, but the fact remains that only "avr" offers best compatibility with libraries following the new scheme.

This is entirely my decision to make.  I value end user experience much more than internal file/directory/naming structure.

PR #55, opened 2015-03-02
Fix Flash and Bus Clock dividers for F_CPU <= 16MHz for TeensyLC

duff2013:
I believe this fixes the bus and flash clocks on the TeensyLC from a previous pull I did: https://github.com/PaulStoffregen/cores/pull/51. I now see where the flash and bus get thier divide value from OUTDIV and OUTDIV4.

I tested Serial1,2,3 and they work from 4-16 MHz but at 2 MHz the baud rate generator can't divide down to most of the standard baud rates for either System or Bus clocks that the serial ports use even though Serial1 has the OSR. Maybe I'm not seeing it but I'll bring this up on the forum.


PaulStoffregen:
thanks

PR #56, opened 2015-03-07
Fix for IDE Menu option "NO USB" @ F_CPU <= 16MHz LC/T3.x

duff2013:
This makes it so you can compile with IDE option "NO USB" at F_CPU <= 16MHz for Teensy 3.x/LC. This is important for LC's small RAM if not using usb at those slower speeds.

### DIFF
--- a/teensy3/usb_inst.cpp
+++ b/teensy3/usb_inst.cpp
@@ -77,6 +77,8 @@ usb_serial_class Serial;
 
 #if defined(USB_SERIAL) || defined(USB_SERIAL_HID)
 usb_serial_class Serial;
+#elif (USB_DISABLED)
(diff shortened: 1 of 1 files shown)


duff2013:
Something I experimented with is to automatically define USB_DISABLE if at the lower speeds by adding this to kentis.h:
```c++
// No need to load USB driver at low speeds
#if ((F_CPU < 2000000) && !defined(USB_DISABLED))
 #define USB_DISABLED
#endif
```
even though it pulls in some code of whatever usb option is selected in the IDE it does skip the actual usb driver code. Maybe this is not the way to do it but I haven't run into any issues yet. I can do pull request if this is something you think would work?

PR #57, opened 2015-03-12
Added support for Single Wire serial mode

davidthings:
Codewise, there are a small handful of changes to implement this.  In the initialization
of the port, several bits are set to tell the UART to do Single Wire mode (LOOPS, RSRC) and the
initial pin direction (TXDIR) is set to 0 (receive).  Then, each time the transmitter starts up,
serial_putchar() and serial_write() for Serial1 and Serial2 only, TXDIR is set to transmit.  In
the ISR, when transmission is ended, TXDIR is returned to the receive direction.

All 36 format settings have not been tested, however, 8N1 has been tested on all three ports in
both Single Wire and non-Single Wire modes.  It is hoped that having a #define remove all trace
of the Single Wire modifications helps to offset the impact of any more complex interactions.

Using this mode is very simple.  Connect the TX pin of your Teensy serial port to the appropriate
pin of another device implementing single wire mode. Then uncomment the SERIAL_SINGLEWIRE_SUPPORT
define in HardwareSerial.h (in ... arduino/hardware/teensy/avr/cores/teensy3).  In your code, open
the correct Serial port with the speed and format desired, adding "_SINGLEWIRE" to the format
specifier.  For example

  Serial1.begin( 1500000, SERIAL_8N1_SINGLEWIRE );

And use the serial port as usual.  All the switching of TX and RX is done automatically and there
is no loop back from TX back into RX again.

[Apologies for the extraneous line changes.  My editor was set to remove trailing whitespace.]


NicoHood:
I have no chance to test the code, but I think this is very useful in case the other pin can be fully accessed and used as I/O pin.

davidthings:
NicoHood, I did not think of this.  You are right, this implementation probably ought to permit the reuse of the RX pin.

I note that pin configuration is done in  ```void serial_begin()``` before ```void serial_format( )``` - where the Single Wire mode is configured - is called.  

```
void serial_begin(uint32_t divisor)
{
   // ...
	CORE_PIN0_CONFIG = PORT_PCR_PE | PORT_PCR_PS | PORT_PCR_PFE | PORT_PCR_MUX(3);
	CORE_PIN1_CONFIG = PORT_PCR_DSE | PORT_PCR_SRE | PORT_PCR_MUX(3);
   //  ...
```

Therefore there would be a little moment at init where the RX pin would act like an RX pin before the Single Wire code shut it down.  Alternatively these two functions could get refactored slightly to make this a bit more elegant.

Let's let Paul decide how he'd like to handle it.

cowboyrushforth:
Hello,

I have managed to get this to compile with my tree of Teensy and it saved the day.  Thank you so much.

Is there plans to integrate this?  I can see it's been awhile for this issue.  Possibility these days there is a better way of going about this?

Cheers / Thanks

TheNetStriker:
I just used this for one of my project where I had to control a device with only one UART wire and it works perfectly! Why was this never merged into the main version?

TheNetStriker:
Are there any updates regarding this pull request? Will this be merged?

PR #66, opened 2015-03-28
use EOSF flag instead of SCNIP to signal scan complete - Teensy 3.x

duff2013:
When using EOSF flag to indicate when the tsi scan is complete you can get rid of delayMicroseconds. I tested with all the tsi pins and it works, the scan delay is almost the same as using delayMicroseconds.


duff2013:
This code as is, is not thread safe because of using delayMicroseconds unless you plan to rework that delay for thread safety. ESOF flag would be better since you can call yield while waiting for the scan to complete.

PR #68, opened 2015-04-14
delay/micros fail until first clock tick

KurtE:
On the Teensy LC - there is an issue that delay was not working at the
beginning of setup.  Turns out that the system tick clock was giving a
real high bogus value, which cause micros to give bad value, until the
first tick at which time the value was reset to a good state.
More details in thread:


KurtE:
Old one, looks like fix is already in

PR #73, opened 2015-05-08
strlen is already in string.h

FrankBoesing:
or, alternativly one could write size_t strlen(const char *s) {return __builtin_strlen(s);} to use the gcc-builtin function.
I think, both are faster than the bytewise search.


FrankBoesing:
Oops  __builtin_strlen(s)  does not work, i have no idea why. The newlib-variant is ok.

PR #77, opened 2015-05-27
Corrected DMA destination size on Teensy LC's destinationCircular met…

kbob:
I think you meant to set the DMA destination size to match the word size.


pedvide:
I've found this bug too, the DMA_DCR_DSIZE(x) in destinationCircular should be the same as in destinationBuffer.

PR #82, opened 2015-06-24
Update begin() - wait for availability

FrankBoesing:
New users often have problems with printing values in setup().
This does not work, because the teensy is too fast, and prints before the usb-connection is ready.
This fix mimics "While (!Serial);" with a additional timeout.

Example: https://forum.pjrc.com/threads/28992-arduino-compatibility-writing-to-serial-monitor-from-setup-does-not-work


FrankBoesing:
..closing old pull-reqs don't worry.

PR #83, opened 2015-06-25
Remove functions already defined in the standard library

GrayShade:
Perhaps I'm supposed to do something else, but with `arm-none-eabi-newlib` I get errors because `utoa` and `itoa` are already defined in `stdlib.h`. This is similar to commit 7d3da62a4817ad5e4d96e4271642d3b294ce0e94.

### DIFF
--- a/teensy3/avr_functions.h
+++ b/teensy3/avr_functions.h
@@ -94,10 +94,14 @@ static inline void eeprom_update_block(const void *buf, void *addr, uint32_t len
 
 char * ultoa(unsigned long val, char *buf, int radix);
 char * ltoa(long val, char *buf, int radix);
+
+#if defined(_NEWLIB_VERSION) && (__NEWLIB__ < 2 || __NEWLIB__ == 2 && __NEWLIB_MINOR__ < 2)
 static inline char * utoa(unsigned int val, char *buf, int radix) __attribute__((always_inline, unused));
 static inline char * utoa(unsigned int val, char *buf, int radix) { return ultoa(val, buf, radix); }
 static inline char * itoa(int val, char *buf, int radix) __attribute__((always_inline, unused));
 static inline char * itoa(int val, char *buf, int radix) { return ltoa(val, buf, radix); }
+#endif
+
 char * dtostrf(float val, int width, unsigned int precision, char *buf);
 
 


PaulStoffregen:
Earlier versions of newlib do not define these.

Any chance you could look into which version added then, and surround this with some sort of #ifdef check for the proper version.

I can't merge this pull request as-is, because it will break compatibility with the version we're still using in Teensyduino.

GrayShade:
I see what you mean. I added a guard, but I didn't test this with Teensyduino.

FrankBoesing:
There is also a bunch of functions for random():
https://sourceware.org/newlib/libc.html#rand

GrayShade:
But that's just a different thing -- and replacing `random()` with `rand()` might lead to trouble for applications that expect the same sequence of values.

The problem with `itoa()` was that the compiler saw a normal function redefined as static and gave an error.

FrankBoesing:
Yes i know. Just noticed that and thought it could be mentioned. There may be potential for optimization.

GrayShade:
I agree, there might still be some low-hanging fruit for size optimization. At some point I noticed that some components take a bit of space in the binary even when they're not in use. Besides the recently-discussed `HWSerial`, the interval timer, tone generator and audio stream classes caught my eye, but I can't recall if the overhead was considerable (it probably wasn't).

And since I'm going off-topic, another avenue which might be worth exploring is `-flto`, which does help (again, I don't have numbers), but is tricky to use. There might have been a better way, but when playing with it I ended up sprinkling `noinline` annotations to prevent stuff from being inlined in the startup section of the binary.

FrankBoesing:
Yes, i played with -flto too, i even tried to compile newlib with -flto but at some point i got errors and i turned to other things .

Last observation for today :-) : with -o3, the resethandler becomes too large with gcc embedded 4.9-2015-q2-update . 

__attribute__ ((section(".startup"),optimize("Os")))
void ResetHandler(void)

is enough - But one can argue if optimize("Os") is the best way. Perhaps that would be a topic for the forum.

Trass3r:
This won't work if you use -std=c++11 instead of -std=gnu+11 as it should be.
See #104.

FrankBoesing:
Arduino uses -std=gnu+11, so switching to -std=c++11 means to to introduce incompatibilities
See platform.txt for AVR

Trass3r:
Doesn't stop anyone from using -std=c++11.

PR #89, opened 2015-10-28
Fix for dtostrf on floats under zero

Defragster:
Per forum
https://forum.pjrc.com/threads/1227-Teensy3-alternative-for-dtostrf()?p=86671&viewfull=1#post86671


Defragster:
AWESOME - glad to help!

PR #92, opened 2015-10-31
Master

Defragster:
Prior course fix of 'e' adjustment resulted in extra pad bytes and final digit round errors. Dozens removed as tested.

In low displayable precision some few remain, only where there was a total failure in pre-existing 1.25 code. as my test app indicated

The fcvt() call perhaps rounding on wrong last digit as used. And where width and precision are similar.


Defragster:
This post for reference shows test results and testing sketch.  Prior posts have info as it was found.
https://forum.pjrc.com/threads/1227-Teensy3-alternative-for-dtostrf()?p=86873&viewfull=1#post86873

Defragster:
ping here too.  This pull updates some of the pre-existing bad cases that were not right. This makes many better with padding and skipped digit errors.  It is not perfect at fixing the fight over precision and width and the string provided from fcvt() that rounds the last digit differently than sprint(), but my tests didn't see any remaining issues that were not there before my changes - per the forum posts.  This includes calls with negative width which rarely worked before and seems unaffected by my changes.

Defragster:
This got done

PR #105, opened 2016-01-03
bug fix, USB system, enable irq only if it was enabled

dfries:
I have a program that's receiving data over USB serial and consuming that data as timers expire (two interrupt routines).  I'm having data corruption, hangs, and other odd problems.  I disabled interrupts in my routines, but I'm calling the usb_serial routines and they unconditionally enable interrupts.  Here are corrected versions of the usb routines that will only enable interrupts if interrupts were previously enabled.  This fixed the problems I was seeing.


dfries:
updated and submitted pull request #361

PR #109, opened 2016-01-16
add optimized memset

FrankBoesing:
tremendous speedup!


FrankBoesing:
Test:

void setup() {
  // put your setup code here, to run once:
  delay(1000);
  Serial.println("Hi!");
  delay(100);
  int m = micros();
  for (int i=0; i< 100; i++)
  memset(mm,0,sizeof(mm));
  m = micros()-m;
  Serial.println(m);
}

void loop() {
  // put your main code here, to run repeatedly:

}

execute this, then apply the patch, restart arduino and execute it again...

PaulStoffregen:
Hmm... running this on Teensy 3.2, I get 1637 us with this new memset().

Without it, I'm getting 1254 us.  Did I do something wrong?

FrankBoesing:
oh.
you're right for "optimized" - maybe i've done something wrong. :-(

But for -Os i still get 1636(with patch from above) / 12834 (original) ms
ok,  1634 is still more than 1254.. seems not to be optimal.

PaulStoffregen:
Added check for -Os mode.
https://github.com/PaulStoffregen/cores/commit/248de3382296a0d2d7c121afe168083d4cc55707

PR #118, opened 2016-03-21
add function eeprom_size()

FrankBoesing:
Useful to know the size !
(And would be very useful to have this  for "TeensyTransfer" )

Would it be possible to add this to 1.28 ??

### DIFF
--- a/teensy3/avr_functions.h
+++ b/teensy3/avr_functions.h
@@ -38,6 +38,7 @@ extern "C" {
 #endif
 
 void eeprom_initialize(void);
+uint32_t eeprom_size(void) __attribute__((const));
 uint8_t eeprom_read_byte(const uint8_t *addr) __attribute__ ((pure));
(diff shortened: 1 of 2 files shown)


PaulStoffregen:
Can you use E2END?

These functions are meant to emulate the avr-libc functions.  On AVR, the standard way to detect the size is E2END defined by avr/eeprom.h.  Teensy supports this way.

FrankBoesing:
E2END does not reflect the real size (it can be smaller than 2KB on T3), so thats the second best solution :)



But i could use it. It's on you.

Or, can we rename that function? If the name is the problem ?

PaulStoffregen:
Oh, I should change things so E2END always shows the actual size.

I'm reluctant to expand the AVR emulation layer with functions which do not exist on real AVR.  I agree eeprom_size() is nice looking.  But long-term I want to avoid confusion as people use and share code both Teensy and Arduino.

Please use E2END.

FrankBoesing:
Ok, thanks :)

PR #125, opened 2016-04-24
Allow the teensy3 library to be used with strict C++11 conformance.

mkende:
The wiring.h file uses the "typeof" construct which is a GNU extension to the language. Under C++11 this is superseeded by the "decltype" keyword.
This change declare "typeof" to be "decltype" whenever the code is compiled with a strict C++11 compiler.


PaulStoffregen:
Looks good

PR #126, opened 2016-04-25
Master

universpharmacie:
Reduce the number of divisions in the micros() function to optimize the
code for KINETISK

### DIFF
--- a/teensy3/pins_teensy.c
+++ b/teensy3/pins_teensy.c
@@ -1086,7 +1086,7 @@ volatile uint32_t systick_millis_count = 0;
 
 uint32_t micros(void)
 {
-	uint32_t count, current, istatus;
+	uint32_t count, current, istatus, reload;
 
 	__disable_irq();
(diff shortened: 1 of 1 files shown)


universpharmacie:
Dear Paul,

while looking through pins_teensy.c, I found that the micros() code could be optimized by using the systick reload value which reduces the number of divisions of F_CPU and makes the code quicker to execute. Tested with Teensy 3.2, code for LC (#ifdef KINETISL) is unchanged.

PaulStoffregen:
Is this really more efficient?

The divisions you're eliminating should be compile-time constants.  The compiler computes then and puts the answer into the code, rather than spending real instructions to do the work at runtime.

universpharmacie:
You are right, comparing the run time for a loop with 10,000 executions of the old and new functions did not show a noticeable difference. Initially, I had in mind that a division could need up to 12 CPU cycles, thus my "wrong" idea about optimizing. Sorry for having wasted your time.

PR #129, opened 2016-05-22
Teensy 2 - Serial1.write() - maybe bypass queue

KurtE:
Add same speedup from Arduino core, that on Serial1.write, check to see
if output queue is empty and data register empty, if so simply stuff the
character out in the output queue, without needing to use interrupt to
start transmission.  Helps speed up fast IO like 1MBS

Also, changed how UCSR1B is updated, to be like Arduino.  In particular
it turns on and off specific bit(s) instead of complete set of register.
Needed if you wish to emulate half duplex with this uart.

### DIFF
--- a/teensy/HardwareSerial.cpp
+++ b/teensy/HardwareSerial.cpp
@@ -146,6 +146,7 @@ void HardwareSerial::write(uint8_t c)
 #endif
 {
 	uint8_t i;
+	uint8_t status;
 
 	if (!(UCSR1B & (1<<TXEN1))) {
 #if ARDUINO >= 100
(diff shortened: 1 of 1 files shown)


PaulStoffregen:
Does this create a race conduction, between writing to the hardware register and setting the "transmitting" flag?

KurtE:
Maybe - Not likely from direct we output this byte and it completes before we set transmitting, but maybe if some other interrupt happens between these two instructions, which gets serviced and by the time it finishes, the character has been output. 

I now put it in a save status register, cli, ... restore status register block and again tested it with my USB2AX like sketch still appears to work and at a high speed...

PaulStoffregen:
ok, looks good

PR #130, opened 2016-05-25
Teensy2 digitalwritefast

KurtE:
For Teensy2...

Add digitalWriteFast, which is a copy of digitalWrite when using built-in const pins else it calls off to digitalWrite.

Add digitalReadFast which simply calls digalRead.

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -390,6 +390,3 @@ update	KEYWORD2
 isEnabled	KEYWORD2
 getFrameCount	KEYWORD2
 XPlaneRef	KEYWORD2
-
-# USB Touchscreen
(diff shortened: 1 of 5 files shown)


PaulStoffregen:
This undoes other recent commits.  I'll manually merge it later today....

KurtE:
Sounds good, most be some changes on your side as I believe I am up to sync with what is checked in.

KurtE:
FYI - I think it picked up more semi- changes due to me syncing up to your changes. 
Only 1 changed file (teensy/core_pins.h)

PaulStoffregen:
manually merged.

Thanks!  :)

PR #135, opened 2016-06-12
add pinDisable(const uint8_t pin)

FrankBoesing:
https://forum.pjrc.com/threads/34800-Are-Teensy3-2-Pins-High-Low-and-Tri-State?p=106555&viewfull=1#post106555


FrankBoesing:
Remember to add the Keyword to ...Arduino\hardware\teensy\avr\keywords.txt, too!

PR #139, opened 2016-06-18
Add support for pins 26 and 27

KurtE:
For Teensy 3.5 add support for 
Serial1.setTX(26)
and 
Serial1.setRX(27)

### DIFF
--- a/teensy3/serial1.c
+++ b/teensy3/serial1.c
@@ -10,10 +10,10 @@
  * permit persons to whom the Software is furnished to do so, subject to
  * the following conditions:
  *
- * 1. The above copyright notice and this permission notice shall be 
+ * 1. The above copyright notice and this permission notice shall be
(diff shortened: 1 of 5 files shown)


KurtE:
As mentioned I the delta comments, I updated all 5 Serialx.c files for the setTX/setRX.
Serial1 has new TX/RX pins on T3.5
Serial2 has fewer TX/RX pins on T3.5

Implemented setTX for all of these chips as to allow you to process the OPENDRAIN option, even though there was only one valid TX pin.

PaulStoffregen:
I made one more little change.  Does this look ok?  https://github.com/PaulStoffregen/cores/commit/5eab8dd0b84f07a8096973ca6add797d25e55c0d

KurtE:
Yes - will do more testing later

PR #140, opened 2016-06-18
Fix for crash @ 48 & 24MHz

FrankBoesing:
RefManual:
"The FlexBus clock frequency must be programmed to be less than or equal to the bus
clock frequency."


PaulStoffregen:
thanks!  :)

PR #141, opened 2016-06-18
180MHZ with USB

FrankBoesing:
Enables USB for 180MHZ - Running from IRC48 Clock

SIM_CLKDIV1_OUTDIV2(2) and SIM_CLKDIV1_OUTDIV4(7) need to be checked if they are ok, , its too late here (after midnight)..


FrankBoesing:
outdated

PR #151, opened 2016-07-03
PulseIn on LC issue

KurtE:
PulseIn may fail on LC as it is using *reg to test the value for on or
off, but on LC register is a bitmask associated with several IO pins.
So split off for LC to use the MASK value for the pin to test high and
low.

Warning did not update timings for this,  which may be necessary as adds
an & in the loop.  However when I tried TLC for 5000us timeout, it timed
out with 5138us, if I decremented the PULSEIN_LOOPS_PER_USEC by one,
then it timed out at 4398... This was at 49mhz.


PaulStoffregen:
Thanks!

PR #158, opened 2016-07-24
T3 5 beta serial6

KurtE:
Add Serial6 support for T3.5 (or is it now T3.6), using LPUART.  Will probably still need some additional updates, but I believe most of the basics are working.  

More details in posting in beta thread. 

Note: I added two separate implementations, for Serial6.  The normal Serial6.cpp is for T3.4->3.5 which will have Uart5, and another one that is for current beta using LPUART


PaulStoffregen:
wow, thanks!  :)

PR #159, opened 2016-07-24
Keycode for Non-US backslash

parsnip42:
Not sure if it belongs in here, but this has been tested and seems to work.


PaulStoffregen:
Thanks.  I manually added this, since it's needed with mask 0xF000 and on both cores.

PR #161, opened 2016-07-24
T35 use sdcard pins

KurtE:
This is optional, but allows users to use the SDCard pins, for other purposes, if they use some adapter to gain access. This Pull allows these pins to be use as GPIO pins 58-63.

Also with related changes in SPI project allows you to use SPI1 on these pins for SCK/MISO/MOSI/CS0-3

Also allows you to use on Serial2 RX/TX/CTS
And Serial4 RX/TX

Also fixed bug in Serial4.setTX which is looking at wrong UART.


FrankBoesing:
Paul, can you please merge these commits? It's very useful for the SDCard pins.

PaulStoffregen:
Ok, let's do it!

PR #166, opened 2016-08-18
T3.6 - If we are going to use HSRUN, read serial number before

KurtE:
If we are runnig at over 120mhz and turn on HSRUN mode, we can not read
in the USB serial number.

This is one attempt to see if we can read it in before we switch to
HSRUN mode.

My quick couple of test appeared to work, but may be cleaner ways

### DIFF
--- a/teensy3/mk20dx128.c
+++ b/teensy3/mk20dx128.c
@@ -842,6 +842,8 @@ void ResetHandler(void)
 	// if we need faster than the crystal, turn on the PLL
    #if defined(__MK66FX1M0__)
     #if F_CPU > 120000000
+	extern void usb_init_serialnumber(void);
+	usb_init_serialnumber();	// see if we can call this before we are in HS mode if we get
 	SMC_PMCTRL = SMC_PMCTRL_RUNM(3); // enter HSRUN mode
 	while (SMC_PMSTAT != SMC_PMSTAT_HSRUN) ; // wait for HSRUN
     #endif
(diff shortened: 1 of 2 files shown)


Defragster:
I did this variation posted on the forum: https://forum.pjrc.com/threads/91-teensy-3-MAC-address?p=113762&viewfull=1#post113762

It uses the location of this - and uses the finding that if the area is read before HSRUN- the value is then present after HSRUN when the read fails.  I don't know if that area may be volatile and something else may wipe it later?

KurtE:
As mentioned by Defragster,  his single file (mk20dx128.c) change appears to work:  Here is the differences from when I edited his in:
```
@@ -842,6 +842,13 @@ void ResetHandler(void)
	// if we need faster than the crystal, turn on the PLL
   #if defined(__MK66FX1M0__)
    #if F_CPU > 120000000

	// Code to get USB serial before entering HSRUN mode
    FTFL_FCCOB0 = 0x41;
	FTFL_FCCOB1 = 0x07;
	FTFL_FSTAT = FTFL_FSTAT_CCIF;
	while (!(FTFL_FSTAT & FTFL_FSTAT_CCIF)) {}

	SMC_PMCTRL = SMC_PMCTRL_RUNM(3); // enter HSRUN mode
	while (SMC_PMSTAT != SMC_PMSTAT_HSRUN) ; // wait for HSRUN
    #endif
```

Defragster:
Pull 176 was taken and makes this obsolete

PaulStoffregen:
fixed by #176

PR #167, opened 2016-08-24
Add A23 and A24 for T3.5 and T3.6 board

KurtE:
I added pins  49 and 50 as analog pins A23 and A24.  I still left in the
special casing for pins 44 and 45 being used for Temp and analog
reference

### DIFF
--- a/teensy3/analog.c
+++ b/teensy3/analog.c
@@ -397,7 +397,7 @@ static const uint8_t channel2sc1a[] = {
 // A18  PTC10  ADC1_SE6b
 // A19  PTC11  ADC1_SE7b
 // A20  PTA17  ADC1_SE17
-	23, 23+128, 26, 18+128  // A21-A22, temp sensor, vref
+	23, 23+128, 10+128, 11+128, 26, 18+128  // A21-A24, temp sensor, vref
 // A21  DAC0  ADC0_SE23
 // A22  DAC1  ADC1_SE23
 };
@@ -452,8 +452,13 @@ int analogRead(uint8_t pin)
 		index = pin - 19;  // 31-39 are A12-A20
 	} else if (pin >= 40 && pin <= 41) {
 		index = pin - 30;  // 40-41 are A10-A11
-	} else if (pin >= 42 && pin <= 45) {
-		index = pin - 21;  // 42-43 are A21-A22, 44 is temp sensor, 45 is vref
+	} else if (pin >= 42 && pin <= 43) {
+		index = pin - 21;  // 42-43 are A21-A22
+	} else if (pin >= 49 && pin <= 50) {
+		index = pin - 26;  // 50 and 51 are A23-A24
+	// Leave compatibility with older where 44 and 45 are Temp and Vref
+	} else if (pin >= 44 && pin <= 45) {
+		index = pin - 19;  // 42-43 are A21-A22, 44 is temp sensor, 45 is vref A25, 26
 	} else {
 		return 0;
 	}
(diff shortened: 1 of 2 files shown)


KurtE:
Note: On my machine I also added A23 and A24 to keywords.txt, but not sure where you pick that up from.

PaulStoffregen:
Let's use 45 & 45 for A23 & A24.

We're still in beta, so breaking compatibility with such a minor feature is still ok.  Let's go with fewer conditional tests for slightly more efficient code.

KurtE:
Sorry, I did not understand.

I understand not needing to keep compatibility with beta test code.   

Could not tell what you are suggesting. Keep the 44, 45 casing as one case statement (not hard) and simply add the 49, 50 at the end?  

Or if I was doing from Scratch, I might have a tendency to use straight lookup table.  if less than 50, just use array index, and maybe check for illegal value like 0xff...

KurtE:
Paul, 
What I am wondering is what your thoughts are on doing this in a similar way that is done in the ADC library.  That is if I define the look up table as passed in pin number to analog channel, it would probably be quicker.  like maybe define the table like:
`static const uint8_t channel2sc1a[]= { // new version, gives directly the sc1a number. 0xff invalid
    5, 14, 8, 9, 13, 12, 6, 7, 15, 4, 3, 19+128, 14+128, 15+128, 		// 0-13, we treat them as A0-A13
    5, 14, 8, 9, 13, 12, 6, 7, 15, 4, 			// 14-23 (A0-A9)
    0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 	// 24-30
    14+128, 15+128, 17, 18,						// 31-34 A12, A13, A14, A15
     4+128, 5+128, 6+128, 7+128, 17+128,		// 35-39
     3, 19+128, 23, 23+128, 26, 18+128,			// 40-45 A10, A11, A21, A22, TEMP, VREF 
     0xff, 0xff, 0xff, 10+128, 11+128 			// 46-48 (not used) A23, A24
};`
Warning I have not done complete verification of this table yet.  Then the beginning code could be as simple as: 
`	if (pin < sizeof(channel2sc1a)) {
		channel = channel2sc1a[pin];
		if (channel == 0xff) {
			return 0;
		}
	} else {
		return 0;
	}
`
Yes there are duplicates at the start of the table and some wasted blanks, but this would be in code space and as I mentioned think this would run reasonably quick.

Thoughts?

KurtE:
Sorry to bother again, but did a quick edit of the code to allow me to compile Analog.c with the straight lookup table like I mentioned and for the T3.6, code size is reduced by 40 bytes.  Test program went from 42248 bytes down to 42208

KurtE:
In this branch I pushed in minor change where I removed one extra test statement, by rearranging lookup table slightly. 

Note: I also pushed up a different branch: https://github.com/KurtE/cores/tree/T3.6-Add-A23-A24-lookup

Where I did do the straight lookup table like ADC library,  It increased code size on some boards. 
I walked through the different Analog pins, on T3.6, T3.2 and TLC.  Did not test out table for T3.0 as I don't have one.  

In that branch only file changed from above is analog.c and there is a #define at top which controls use table or old code.  I mainly pushed it up in case you wish to take a look.

KurtE:
As a possible extension to this, I defined A25 and A26 for T3.5 to use the D+ and D- pins from the USB host connector.  I have done some testing with it and does look promising. 

Code is up on the branch: https://github.com/KurtE/cores/tree/T3.5-Also-Add-A25-A26
If you are interested you can probably easily pull from there or I can issue different Pull request

PaulStoffregen:
> As a possible extension to this, I defined A25 and A26 for T3.5 to use the D+ and D- pins from the USB host connector. I have done some testing with it and does look promising. 

I've added A25 & A26 support with this commit.
https://github.com/PaulStoffregen/cores/commit/28d7ba477e00410b68e2d9c7900097e5af9dfae6

Looks like there's probably a hardware bug impacting the polarity of A26...

PR #169, opened 2016-08-31
Fix Serial6 for Teensy 3.5

KurtE:
Needed to add new Define for clock gate

Plus defines to use the new clock gate, plus correct case name for the
serial6 isr to match what is in system isr table


PaulStoffregen:
Thanks!

PR #171, opened 2016-09-03
Add System Register File and VBAT Register File

FrankBoesing:


PaulStoffregen:
These names are very short.  Maybe they could conflict with names in other headers?

Looks like Freescale doesn't really define register names for these?

On the register file, maybe it should be an array with defined size?  Does it need to be volatile?

FrankBoesing:
Hi Paul,

Freescale uses
```#define RFSYS_BASE (0x40041000u)```

We can use

```SYSREGISTERFILE    ((uint8_t *)0x40041000) // System Register File```

or VBATREGISTERFILE.

This definition above allows to use it as array.

What do you think ?



Am 03.09.2016 um 12:31 schrieb Paul Stoffregen:

PR #173, opened 2016-09-07
T3.5/3.6 - SPI has additional Miso/Mosi/SCK pins

KurtE:
Added in the additional SPI pins
MISO 39
MOSI 28
SCK 27

Tested with my ILI9341_t3n (SPIN) version.  Will also
Add to ILI9341_t3.


KurtE:
Earlier when I implemented SPI1 for T3.5/3.6, I had the default SCK1 as pin 20 as I believe that this is the pin used on TLC.  But I see the card showing pin 32 as the default so add delta to switch to 32 as the default pin

PR #176, opened 2016-09-29
HSRUN drop/raise to allow Ser# read for USB :: Defragster patch 1

Defragster:
This #if code is T_3.6 specific to F_CPU>120000000
As tested in my EEPROM write at HSRUN, with interrupts off ( they already are ) dropping HSRUN for a short time of simple xfer like EEPROM write or the current CORES serial # register read.


Defragster:
Cool.

PR #177, opened 2016-09-29
EEPROM write on T_3.6 under HSRUN - Drop and Raise

Defragster:


Defragster:
This code checks and conditionally restores 'extraneous' things:

- **HSRUN** : if large # of eeprom writes are needed HSRUN could be dropped by the caller (prior scheme)
- **interrupts** : same as above - or if a simple write is done when they were already off when called
- 

This assures no surprises at run time and in timed testing is minor compared to EEPROM write time.

Defragster:
see:: https://forum.pjrc.com/threads/36808-EEPROM-writing-on-T_3-6-in-HSRUN-mode?p=117249&viewfull=1#post117249

Minor cleanup and paranoia - no problems were seen.

tni:
Your code is not ISR / thread safe. See code comment.

teensy3/eeprom.c
```diff
@@ -183,18 +186,68 @@ static void flexram_wait(void)
...
+void c_enable_irq( void ){
+	if ( c_intrestore ) {
+		c_intrestore =0;
+		__enable_irq( );
+	}
+}
+void c_disable_irq( void ){
+	if ( !__get_primask() ) { // Returns 0 if they are enabled, or non-zero if disabled 
+		c_intrestore = 1;
+		__disable_irq( );
```
Defragster: Made this change - pending review - thanks for feedback tni

Defragster:
Updated order of Interrupt Restore flag setting to better track and enable_irq on exit if interrupted.

PaulStoffregen:
I'm a little concerned about whether this is really correct, but I'm going to merge it now and ask questions later.....

PaulStoffregen:
Ok, my main question is about the clock multiplier.  As nearly as I can tell, this code is turning off HSRUN mode, but leaving the clock set above 120 MHz.  Did I miss something where it deals with first reducing the clock speed before disabling HSRUN?

teensy3/eeprom.c
```diff
@@ -183,18 +186,68 @@ static void flexram_wait(void)
...
+	}
+}
+static volatile uint8_t restore_hsrun = 0;
+static void hsrun_off(void)
+{
+	if (SMC_PMSTAT == SMC_PMSTAT_HSRUN) {
+		c_disable_irq( ); // Turn off interrupts for the DURATION !!!!
+		SMC_PMCTRL = SMC_PMCTRL_RUNM(0); // exit HSRUN mode
+		while (SMC_PMSTAT == SMC_PMSTAT_HSRUN) delayMicroseconds(2); // wait for !HSRUN
+		delayMicroseconds(100);
```
duff2013: Also is this delay needed?

duff2013:
I don't see it dealing with the reducing cpu speed but maybe it doesn't matter? Also why is delayMicroseconds used when waiting for it to switch off/on HSRun Mode?

Defragster:
To summarize what is posted later on the forum thread: https://forum.pjrc.com/threads/36808-EEPROM-writing-on-T_3-6-in-HSRUN-mode?p=117249&viewfull=1#post117249

**When HSRUN is Dropped :: The clock speed is NOT dropped - No frequencies are modified.**
- that is why it seemed essential to have ZERO interrupts to limit activity to EEPROM writes
- same applies to Serial # transfer and EEPROM writes - no speed changes

**DELAY added to IRQ and HSRUN changes - I had no delays and all was well** 
- Net throughput at 240 with delays is still typically 2x faster than at 120
- From delay 0 to toward 100 in line #217 there was no measurable loss in throughput.  It seems the processor is already pausing to adjust to the HSRUN transition
- History:: The line #216 delay helped when I was changing speeds to restore USB

**I tested without delays - then added and testing there now - if there is some edge problem with internal operation changes for HSRUN it would be best not be on their edges. No net loss in performance on efficient block writes - all 4K byte writes <10% change**

Defragster:
I brought Paul's updated github CORES/teensy3 to my machine and rebuilt and see the HSRUN EEPROM write code compiling and working.

And TYQT shows the expected Serial number on power up.

PaulStoffregen:
I'm working right now on functions to disable and enable HSRUN mode, with clock speed changes for all combinations of F_CPU and F_BUS.....

FrankBoesing:
Clock speed changes in a running sketch: This interrupts _all_ communications, timers, clocks. 
Audi, Uart, .....
Are you sure ? Or did i miss something ?

Defragster:
I had it working 180<>120 MHz [post 38 9/13/16] and got USB back online with 10-100ms delay adjustments to init timing - sooner was a fail. Clocks will slip - and there is some time involved of course waiting for the EEPROM writes to complete.  I expect saying interrupts need to be off the same isn't needed - as going anywhere will likely result in failure with clocks altered. But without significant delays the USB would not restore - you may know a cleaner way to do that.

I was doing it from user code because a group of writes would happen under a single longer drop in speed/hsrun.  Multiple writes will be multiple drop/restore cycles - and how I did it had big overhead.

Without your feedback - I went with HSRUN drop only and it has been great the last month. My PROTO_K66 and beta T_3.6 still running my tests.

Defragster:
Indeed Frank - Hoping Paul's inside knowledge gives him advantage over my efforts.  USB alone was a challenge to restore - even with forced flush on entry. It was block style assuming user would write all that was needed then reset the processor.

Defragster:
posted code with speed (180<>120) change was 9/10: https://forum.pjrc.com/threads/36808-EEPROM-writing-on-T_3-6-in-HSRUN-mode?p=114864&viewfull=1#post114864

post 8 was switch to HSRUN toggle only.

It shows flush and did flush and delay(100) before entry dropping speed.

On speed restore some combination of this was needed to have USB functional - as I did it::
`  __enable_irq( );
  delay(300); // Need this and Windows sees the port return
  usb_init();
  delay(1000);      // for USB setup - delay() added
  Serial.flush();   // for USB setup - output appears out of order - .flush() added
 `
I did minimal refinements after that before bailing on it .

PaulStoffregen:
Ok, I've just committed the HSRUN functions.  https://github.com/PaulStoffregen/cores/commit/1bdf4faa90eb5a1af8fa6c7c06e4855f329500d7

Here's a test program to see their impact, on USB, Serial1, Serial3, and execution speed (watch the LED or use an oscilloscope with 0.5/sec time base).

    void setup() {
      pinMode(13, OUTPUT);
      Serial1.begin(115200);
      Serial3.begin(115200);
    }

    void loop() {
      static int count=0;
      kinetis_hsrun_disable();
      Serial.print("on count=");
      Serial.println(count);
      Serial1.print("on count=");
      Serial1.println(count);
      Serial3.print("on count=");
      Serial3.println(count);
      digitalWrite(13, HIGH);
      delay(500);
      kinetis_hsrun_enable();
      Serial.print("off count=");
      Serial.println(count);
      Serial1.print("off count=");
      Serial1.println(count);
      Serial3.print("off count=");
      Serial3.println(count);
      digitalWrite(13, LOW);
      delay(500);
      count++;
    }

Defragster:
well that is anti-climatic . . .  just change the one clock . . . it works.

I reinstalled 1.31b1 and replaced the two edited files.  USB works with sketch above.

Then the EEPROM test sketch used writing various ways works with edited eeprom.c I edited:
`static void hsrun_off(void)
{	kinetis_hsrun_disable(); }
static void hsrun_on(void)
{	kinetis_hsrun_enable(); }
`
So deleting my overhead there but text swapping the _on and _off calls for the _enable _disable should be complete.

PaulStoffregen:
Yes, I like anticlamactic for low-level code.  ;)

I've committed a change to use these functions in eeprom.c: https://github.com/PaulStoffregen/cores/commit/4797fce596a165f4f790c64af82b552d1443cd6d

So far I've done only basic testing.  Any chance you could try it with those performance benchmarks?

Defragster:
Indeed - minimal change IN SPEC - for low level code is very likeable. 

In general as I saw before time wise is a mixed bag - 120 MHZ was/is from twice as fast to slightly or 10 times slower than 240 MHz.
The NEW 4K byte writes twice as fast (OLD more calls more overhead - or just at 120  alone)- the NEW block writes typically much slower (OLD more work at higher speed). The NEW code splits some difference running at 240 up until the slowing for write.

My time tests were general enough to see when I made some change. They were not an exact programmatic benchmark. The fact that some are 10 times faster than running clocked at 120 MHz is actually what concerned me.

The fact that it works at all - with the limits of the MPU design for <=120 MHZ EEPROM writes - while it is doing what needs to be done to function long term is the best it can offer.

The ones that seem 10 times slower I think are .put of 128 and 256 bytes.  That seems to be because the .put code does single byte writes in the current code,  If you read this post by tni this 'pending' code could correct that: https://forum.pjrc.com/threads/36808-EEPROM-writing-on-T_3-6-in-HSRUN-mode?p=119008&viewfull=1#post119008 where IIRC a write_block is done for that .put case.

Defragster:
Paul:
I tested at 6 speeds over 120 to work against my write test sketch.

USB fails at 120 MHz.  Does this (bottom) piece of code set HSRUN at ALL F_CPU speeds on T_3.6! The return(0) is #ifdef. All of this code should #ifdef out at <= F_CPU==120000000, not just 

`#ifdef HAS_KINETIS_HSRUN
extern int kinetis_hsrun_disable(void);
extern int kinetis_hsrun_enable(void);
#else
`

`int kinetis_hsrun_enable(void)
{
	if (SMC_PMSTAT == SMC_PMSTAT_RUN) {
		// Turn HSRUN mode on
		**SMC_PMCTRL = SMC_PMCTRL_RUNM(3);**
		while (SMC_PMSTAT != SMC_PMSTAT_HSRUN) ; // wait`

FrankBoesing:
One could think that we don't need HSRUN at 144MHz at all.. T3.2/3.5 don't need it, too ?

Defragster:
Paul - run your posted sketch at 120 MHz against your timing scope and you should see what I see - watch pin 13.  The blink time is messed up.

In my sample there is an EEPROM write once a second in "void systick_isr() {" and I lose all sense of time and USB at 120 MHz.

Frank - the Processor has a speed restriction over 120 MHz.  If you set the clocks over 120 MHz - they will not run until HSRUN is entered.  I posted that in my EEPROM thread.

PaulStoffregen:
How's this?  https://github.com/PaulStoffregen/cores/commit/809482c72e21aa75c7267e77c09e37b268e22d0e

PaulStoffregen:
> T3.2/3.5 don't need it, too ?

HAS_KINETIS_HSRUN is only defined on Teensy 3.6.

FrankBoesing:
You mis-understood me. I meant: They work without HSRUN. I guess, it's the same for 3.6 (for 144MHz)
Perhaps enable HSRUN- switch to144, disable HSRUN during initialization. But, yes, that's an other topic...

Defragster:
Paul:  Looks good.
Swapped those two files - compiles and runs 96, 120 and 240 MHz

Frank - indeed I understood you meant T_3.6 only - but since the processor blocks speeds over 120 without HSRUN - I suspect they saw a reason.  My extrapolated inference (based only on Paul's one post about multiple fabrication types) is that the non 5V tolerance of the T_3.6 is due to mixed types of construction within the processor. I have not seen my beta T_3.5 run reliably over 120 MHz.

PaulStoffregen:
I've heard at least one very reliable report of 144 MHz on Teensy 3.2 appearing to work, but crashing randomly on the time scale of days or weeks.  Reducing to 120 MHz solved the problem.

FrankBoesing:
Well - i just tried it. Its just adding two lines to mk20dx12 :)
Works. Same for 168 MHz.

PaulStoffregen:
I'd be willing to change the "120000000" to a symbol that can be overridden from the command line, like F_BUS overclocking.

    #if defined(HAS_KINETIS_HSRUN) && F_CPU > 120000000

Even if it works, 144 MHz without HSRUN is overclocking.  See Table 6-1 on page 120 of the K66 reference manual.

For the upcoming 1.31 release, I really want to get these new boards as stable as possible.  I'm also happy to support overclocking, perhaps even with more options in boards.txt and Arduino's menus, but it needs to be selected by the user with a clear indication it's overclocking.

Defragster:
Frank - Not surprised - All my EEPROM code ran fine without HSRUN even at 240 MHz - and IIRC I started without interrupts disabled. It's usually okay to run with scissors too.

PaulStoffregen:
Also, on the matter of overclocking, I'm considering adding menu options for faster F_BUS speeds.  The main issue is how to explain this in a way people understand, which fits into the small space of the menu.

FrankBoesing:
Okay :) 
Off-topic: What *really* would be great: An file that gets #included for every compile,  - that would allow to fix the settings (like -D... at commandline) for projects (override the ones from the IDE - just by writing the defines to the file. Just as Idea that i want to mention, no need to discuss it here, further :)

Issue #181, opened 2016-10-09
Pin interrupt dispatching is not thread/interrupt safe


tni:
The port interrupt pin dispatch code
`	if ((isfr & CORE_PIN1_BITMASK) && intFunc[1]) intFunc[1]();`
compiles down to the following:
```
...
  16:	03a1      	lsls	r1, r4, #14
  18:	d504      	bpl.n	24 <portb_interrupt+0x24>
  1a:	4b1a      	ldr	r3, [pc, #104]	; (84 <portb_interrupt+0x84>)
  1c:	685a      	ldr	r2, [r3, #4]
  1e:	b10a      	cbz	r2, 24 <portb_interrupt+0x24>
  20:	685b      	ldr	r3, [r3, #4]
  22:	4798      	blx	r3
...
```

`ldr	r2, [r3, #4]` is a load from the intFunc[] table, loading the pin interrupt address. A NULL check is done, then the intFunc[] table load is performed again and the pointer from the table is called.

A higher priority ISR could be called between the two loads from intFunc[]. If that ISR uses detachInterrupt for the pin, NULL will be called.

FrankBoesing:
In theory, you are right.

But the chances that somebody calls detachInterrupt inside an interrupt 
is quite low, i think.

On the other hand, if you have better code _that is not slower_ , 
repairing this would be good. Maybe in assembler to make sure that it 
works for every optimization level and future GCC versions ?

tni:
There are some uses cases. This is an issue that would be almost impossible to track down, since would occur extremely rarely. The Teensy would just hang every few weeks or months.

PaulStoffregen:
This was fixed by #181, right?

tni:
Fixed by #182.

PR #182, opened 2016-10-09
Fix issue #181. Improved pin interrupt dispatch performance.

tni:
Instead of putting a NULL pointer in the pin interrupt target for unused pin interrupts, a dummy function that simply returns is used. This way, no check is required before dispatching. Normally, unused pins will have their interrupt masked and this dummy function will only be called in very unusual circumstances.

Each pin interrupt dispatch is reduced from 7 instructions down to 4.

Previous:
```
  16:	03a1      	lsls	r1, r4, #14
  18:	d504      	bpl.n	24 <portb_interrupt+0x24>
  1a:	4b1a      	ldr	r3, [pc, #104]	; (84 <portb_interrupt+0x84>)
  1c:	685a      	ldr	r2, [r3, #4]
  1e:	b10a      	cbz	r2, 24 <portb_interrupt+0x24>
  20:	685b      	ldr	r3, [r3, #4]
  22:	4798      	blx	r3
```
New:
```
  12:	03a8      	lsls	r0, r5, #14
  14:	6863      	ldr	r3, [r4, #4]
  16:	d500      	bpl.n	1a <portb_interrupt+0x1a>
  18:	4798      	blx	r3
```

### DIFF
--- a/teensy3/core_pins.h
+++ b/teensy3/core_pins.h
@@ -126,6 +126,29 @@
 #define CORE_NUM_PWM            22
 #endif
 
+// These MAX_PIN_PORTx values have the highest Kinetis pin index
+// that is used for a given port.
+#if defined(__MK20DX128__) || defined(__MK20DX256__)
+#define CORE_MAX_PIN_PORTA        13
+#define CORE_MAX_PIN_PORTB        19
+#define CORE_MAX_PIN_PORTC        11
+#define CORE_MAX_PIN_PORTD         7
+#define CORE_MAX_PIN_PORTE         1
+#elif defined(__MKL26Z64__)
+#define CORE_MAX_PIN_PORTA         2
+#define CORE_MAX_PIN_PORTB        17
+#define CORE_MAX_PIN_PORTC         7
+#define CORE_MAX_PIN_PORTD         7
+#define CORE_MAX_PIN_PORTE        30
+#elif defined(__MK64FX512__) || defined(__MK66FX1M0__)
+#define CORE_MAX_PIN_PORTA        29
+#define CORE_MAX_PIN_PORTB        23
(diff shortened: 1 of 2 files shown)


tni:
This second commit has fast dispatching based on CTZ for Kinetis K. It doesn't make sense for Kinetis L (Cortex M0), since there is no hardware CLZ.

Cortex M4 has a single-cycle rbit instruction to reverse the bits, which is used in conjunction with CLZ for a fast CTZ. Using CTZ saves several instructions over using CLZ-based ISR table indexing.

The 'normal' path when only 1 pin has a pending interrupt is 14 instructions:
```
   0:	b538      	push	{r3, r4, r5, lr}
   2:	4b12      	ldr	r3, [pc, #72]	; (4c <port_A_isr+0x4c>)
   4:	681c      	ldr	r4, [r3, #0]
   6:	601c      	str	r4, [r3, #0]
   8:	b1f4      	cbz	r4, 48 <port_A_isr+0x48>
   a:	4d11      	ldr	r5, [pc, #68]	; (50 <port_A_isr+0x50>)
   c:	fa94 f3a4 	rbit	r3, r4
  10:	fab3 f383 	clz	r3, r3
  14:	f855 3023 	ldr.w	r3, [r5, r3, lsl #2]
  18:	4798      	blx	r3
  1a:	1e63      	subs	r3, r4, #1
  1c:	401c      	ands	r4, r3
  1e:	d013      	beq.n	48 <port_A_isr+0x48>
...
  48:	bd38      	pop	{r3, r4, r5, pc}
```

PaulStoffregen:
Wow, thanks!  :)

PaulStoffregen:
This is something I had intended to do for years, but never got around to it.

PR #183, opened 2016-10-10
Add memory_total

FrankBoesing:
Its sometimes very useful to know how much memory was allocated - it
allows, together with memory_used, how many blocks are free. Esp. useful
for queue-like objects (I'm working on a MP3 integration to the
Audio-Library - it needs to know whoe much blocks are free before trying
to decode aMP3-Frame.)


FrankBoesing:
Ok, i use an other solution now. Let's close this.

Issue #122, opened 2016-04-11
GCC 5.3 random()


FrankBoesing:
Just installed GCC 5.3 and got the following error:

C:\Arduino\hardware\teensy\avr\cores\teensy3/WProgram.h:51:21: error: ambiguating new declaration of 'uint32_t random()'

 uint32_t random(void);

                     ^

In file included from c:\arduino\hardware\tools\arm5\arm-none-eabi\include\stdlib.h:11:0,

                 from C:\Arduino\hardware\teensy\avr\cores\teensy3/WProgram.h:4,

                 from C:\Arduino\hardware\teensy\avr\cores\teensy3/Arduino.h:1,

                 from f:\build4def553c7c5cc6638db6fb6a30be3b17.tmp\sketch\coremark.h:27,

                 from f:\build4def553c7c5cc6638db6fb6a30be3b17.tmp\sketch\core_list_join.cpp:20:

c:\arduino\hardware\tools\arm5\arm-none-eabi\include\stdlib.h:231:6: note: old declaration 'long int random()'

 long _EXFUN(random,(_VOID));

      ^

FrankBoesing:
In addition, for -Og and -Os, fcvt() and fcvtf() are not extistant (used in nonstd.c by dtostrf()

tulth:
In teensy3/WProgram.h , I commented out this declaration:
 uint32_t random(void);
As a work around.  

WMath seems to conflict with arm-none-eabi stdlib.h

Ideas?  Rename random functions in WProgram.h?
Make the renaming of WMath random optional with it defaulted off?

PR #187, opened 2016-10-26
Fix compilation error with recent newlib/gcc versions because of random()

ashtuchkin:
`long random()` was added to stdlib.h in newlib project used by gcc arm toolchains in this [commit](https://github.com/bminor/newlib/commit/ecf453f9635fb278cff4d4bae21a1e249313b817). 
This gives compilation error when compiling with [gnu-arm-embedded](https://launchpad.net/gcc-arm-embedded) toolchain starting from 5_3-2016q1 version.

The diff changes signature to signed int, so that there's no ambiguity. Linker will then choose the version of this function from WMath.h (because it'll be from the same module).

This change is safe to do because the internal state/seed is a signed int, so we won't overflow.

Fixes PaulStoffregen/cores#122.

### DIFF
--- a/teensy3/WMath.cpp
+++ b/teensy3/WMath.cpp
@@ -12,7 +12,7 @@ void srandom(uint32_t newseed)
 	seed = newseed;
 }
 
-uint32_t random(void)
+int32_t random(void)
 {
 	int32_t hi, lo, x;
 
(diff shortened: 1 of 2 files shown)


PaulStoffregen:
I wonder if this will break any Arduino sketches which might depend on the unsigned type?

Going to merge anyway, and wait to see if anyone reports errors....

ashtuchkin:
Thanks Paul!

PR #190, opened 2016-10-30
allow setting buffersizes

FrankBoesing:
If this is ok, i can provide this for the other uarts, too.
+ It is possible to deactivate them completely via #define
(saves memory esp. on LC)

### DIFF
--- a/teensy3/serial1.c
+++ b/teensy3/serial1.c
@@ -35,9 +35,12 @@
 ////////////////////////////////////////////////////////////////
 // Tunable parameters (relatively safe to edit these numbers)
 ////////////////////////////////////////////////////////////////
-
-#define TX_BUFFER_SIZE     64 // number of outgoing bytes to buffer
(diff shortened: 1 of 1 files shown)


PaulStoffregen:
Let's make this SERIAL1_TX_BUFFER_SIZE and SERIAL1_RX_BUFFER_SIZE.

PaulStoffregen:
Also considering a proposal (not yet public) for an extension to add a 2nd buffer at runtime.  That will add overhead, so still undecided if I like the idea....

FrankBoesing:
No problem - want it for all serials ? With disable ?
Hm.. a second buffer would be a bit slower. But having that at runtime, would be more easy to use..

PaulStoffregen:
Sure, let's do it for all 6.  Can you amend this PR?

FrankBoesing:
Don't know how .)  Easier for me is to close it :) I make a new one.

PR #192, opened 2016-10-30
SERIAL1_DISABLE..SERIAL6_DISABLE

FrankBoesing:
Can save flash & ram, increase speed of loop() and others due to less
available() -calls


PaulStoffregen:
This is not the way I want to solve this problem.

Yes, I know the proper way is a lot more work.  But it is possible, with a significant restructuring of which code is in which files.  Arduino accomplishes this with weak symbols, but they only result in the code not being linked into the final executable if all the other code accessing the weak alias is in the same file.

PR #193, opened 2016-11-04
infinite loop in dtostrf()

replicater81:
We have infinite loop in dtostrf(), when **val** is NaN or infinity value. Added some code to fix that.


PaulStoffregen:
Fixed by https://github.com/PaulStoffregen/cores/commit/eae654412c8930cf89e93053e94293c35cc0476a

PR #195, opened 2016-11-07
Set RTS Watermarks dependend of rx-buffersize

FrankBoesing:
for 64, it results in exactly th old values (40/26)

### DIFF
--- a/teensy3/serial1.c
+++ b/teensy3/serial1.c
@@ -42,8 +42,8 @@
 #ifndef SERIAL1_RX_BUFFER_SIZE
 #define SERIAL1_RX_BUFFER_SIZE     64 // number of incoming bytes to buffer
 #endif
-#define RTS_HIGH_WATERMARK 40 // RTS requests sender to pause
-#define RTS_LOW_WATERMARK  26 // RTS allows sender to resume
+#define RTS_HIGH_WATERMARK (SERIAL1_RX_BUFFER_SIZE-24) // RTS requests sender to pause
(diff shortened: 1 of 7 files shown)


PaulStoffregen:
This doesn't seem like a good idea for the high watermark.  It makes nearly all of a larger buffer unusable.

FrankBoesing:
? Did you test this ?  :-)

With my change: Assuming a Size of 1024, high-watermark is 1000. The sender gets a request to pause when the buffer is filled with 1000 Bytes

Without: high-watermark is 24. The high.watermark tells the sender "please pause" when the buffer is filled with 24 bytes.

This behaviour refelcts exactly the comment in your code:
"RTS_HIGH_WATERMARK 40 // RTS requests sender to pause"

PaulStoffregen:
ok, I need to look at this more carefully.....

FrankBoesing:
Paul, my patch works..

PaulStoffregen:
Will look at this (and many others soon).  Business stuff has been keeping me from dev work lately.  We had an employee leave and just hired a new person who's coming up to speed after we were short-staffed for 2 weeks.  Robin was also out of town for several days.  Normally PJRC runs smoothly with 4 people, but when short labor-wise, things have to get done which really restricts my dev time.  Getting 1.31 released was top priority, and at the moment I'm trying to wrap up almost 2 dozen Kickstarter rewards that were lost by the postal service.  Please be patient.  I can do a lot, but not everything all at once.

PR #196, opened 2016-11-07
Making this delay(400) split gets T_3.0 and 3.6 online in 405ms

Defragster:
The delay(400) before usb_init() results in USB first Serial.print in about 800ms with newer faster TYQT - On Windows 10.

Making this change keeps the setup() entry time the same ( for i2c etc ) - but USB is now online at 405ms.

### DIFF
--- a/teensy3/pins_teensy.c
+++ b/teensy3/pins_teensy.c
@@ -578,8 +578,9 @@ void _init_Teensyduino_internal_(void)
 	// for background about this startup delay, please see these conversations
 	// https://forum.pjrc.com/threads/36606-startup-time-(400ms)?p=113980&viewfull=1#post113980
 	// https://forum.pjrc.com/threads/31290-Teensey-3-2-Teensey-Loader-1-24-Issues?p=87273&viewfull=1#post87273
-	delay(400);
+	delay(40);
 	usb_init();
+	delay(360);
 }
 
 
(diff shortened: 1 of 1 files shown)


PaulStoffregen:
This delay is needed for some USB hubs.  :(

Defragster:
Bummer - I knew there were timing issues but not related to what :(
If you made the used value 400 a #define it could be swapped out at compile time with the command line / Franks defs.h for the IDE when usage allowed? It would be an obscure 'advanced' item (what isn't :) - but easier than hacking cores with each new TD.

Defragster:
One other note - if this delay is USB specific - having it drop out on non-USB would be nice, but of course optional as some devices would panic if started sooner. So the #define for optional removal would be useful.

Defragster:
Paul: On "Those" USB hubs - is the full delay(400) needed before usb_init() - or just having a total delay of 400 as I do in my request?  Delay before usb_init() won't affect the HUB as the Teensy is a non-device?

I played with the before and after split and settled on some small delay before because less didn't result in any faster Serial online - no matter the second number.

And no matter the second number my hubs through Win 10 rarely hit under 400 - but still come online just fine.  So I stuck with your 'tested' 400 and just made the big part of it after - which meant that the Teensy only waits 2-19 ms in the while( !Serial ) - except for the rare 635 when I do TYQT reset too quickly and I suppose Windows gets held up.

In any case - if those were made #define values [ even if (400) before and (0) after ] - an expert user could modify to fit their system.

Defragster:
Given that delay() calls yield - shouldn't delay(400); be replaced with delayMicroseconds(400000); and USB and otherstuff is offline so far?

Where my edit is in pins_teensy.c
`	//delay(400);
	delayMicroseconds(400000);
	usb_init();`

FrankBoesing:
For my DELL-Screen inbuilt-usb.hub, a delay(400) is not enough

Defragster:
Frank I'll email you with something to try so you can tell me how it fails.

PaulStoffregen:
Ok, let's give this a try.  Really, really hoping we can get more people to help beta test......
https://github.com/PaulStoffregen/cores/commit/da92eda747fe7b65d7898401410c221b385b3c92

PR #200, opened 2016-12-04
AudioStream.h: Allow inclusion from assembler .S files

FrankBoesing:
This way, we can use it (i.e.) in memcpy_audio.S
(Pullrequest follows)


FrankBoesing:
this is for https://github.com/PaulStoffregen/Audio/pull/214

PR #202, opened 2016-12-18
Fix: write returns zero

FrankBoesing:
https://forum.pjrc.com/threads/40893-Incorrect-return-for-Serial-write()-on-Teensy-3-6?p=127732#post127732

I'm not sure if this fix has some side-effects - how will other functions react when it does not return zero anymore ?


FrankBoesing:
On the other hand.. if this breaks something, the bug is NOT here :)
So i guess, it will work.

PR #206, opened 2017-01-13
fix limit checks readString/readStringUntil

FrankBoesing:
https://forum.pjrc.com/threads/41282-Strange-looking-limit-checks-in-Teensy3-Version-of-Stream-cpp


PaulStoffregen:
thanks :)

PR #210, opened 2017-01-22
__cxa_guard_acquire - match declarations in mk20dx128.c

FrankBoesing:
- got a compiler - warning with GCC 6 and LTO
This fixes the warning.
I don't know wether GCC5 prints the warning, too.

### DIFF
--- a/teensy3/new.h
+++ b/teensy3/new.h
@@ -16,8 +16,8 @@ void operator delete[](void * ptr);
 
 __extension__ typedef int __guard __attribute__((mode (__DI__)));
 
-extern "C" int __cxa_guard_acquire(__guard *);
-extern "C" void __cxa_guard_release (__guard *);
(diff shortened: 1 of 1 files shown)


tni:
This is the wrong fix. The ARM ABI prescribes the signature of __cxa_guard_acquire and it's supposed to be: 
int __cxa_guard_acquire(int *gv)

The toolchain cxxabi.h header has it as:
int  __cxa_guard_acquire(__guard*);

So mk20dx128.c should be fixed instead.

PaulStoffregen:
Any idea what header ought to be included to get the definition of "__guard"?  When compiling .cpp including stdlib.h does it, but doesn't seem to work when compiling .c.

tni:
It looks like "__guard" is only defined in the C++ headers (but they should work in C mode). It seems you would need to add the C++ header location to the include path - probably not worth the hassle and potential issues.

PR #217, opened 2017-02-12
Fix to dtostrf for small numbers

LAtimes2:
This fixes a couple of items when converting small numbers. Seems to be a bug in fcvtf, which this is a workaround. One issue is with trailing zeros, other is with rounding. See this post for some examples.

https://forum.pjrc.com/threads/1227-Teensy3-alternative-for-dtostrf()?p=104967&viewfull=1#post104967

Here is a sketch that shows the errors and verifies fixes after incorporation.

```
void setup() {
  // put your setup code here, to run once:
  char str[100];

  float values[] = {
    // leading/trailing 0 issue
     0.01, 0,   0.01, 1,    0.01, 2,   0.01, 3,   0.01, 4,
    0.001, 0,  0.001, 1,   0.001, 2,  0.001, 3,  0.001, 4,
    // rounding issue
    0.0099, 0,   0.0099, 1,    0.0099, 2,   0.0099, 3,   0.0099, 4,
    // these pass - regression tests
    1.23, 0,   1.23, 1,    1.23, 2,   1.23, 3,   1.23, 4,
    9.87, 0,   9.87, 1,    9.87, 2,   9.87, 3,   9.87, 4,
    -12.34567, 0,   -12.34567, 1,    -12.34567, 2,   -12.34567, 3,   -12.34567, 4,
    };

  Serial.begin(9600);
  while (!Serial);
  Serial.println("hello");
  for (int i=0; i < sizeof(values)/sizeof(values[0]); i=i+2)
  {
    dtostrf(values[i], values[i+1] + 2, values[i+1], str);
    // print expected followed by actual
    Serial.print(values[i], values[i+1]); Serial.print("|"); Serial.print(str); Serial.print("|"); Serial.println();
  }
  Serial.println("done");

}

void loop() {
  // put your main code here, to run repeatedly:

}
```

### DIFF
--- a/teensy3/nonstd.c
+++ b/teensy3/nonstd.c
@@ -126,15 +126,58 @@ char * dtostrf(float val, int width, unsigned int precision, char *buf)
 	}
 
 	s = fcvtf(val, precision, &decpt, &sign);
+
+	// if only 1 digit in output
 	if (precision == 0 && decpt == 0) {
+		// round and move decimal point
 		s = (*s < '5') ? "0" : "1";
-		reqd = 1;
-	} else {
-		reqd = strlen(s);
-		if (reqd > decpt) reqd++;
-		if (decpt == 0) reqd++;
+		decpt++;
 	}
+
+	// if all zeros, limit to precision
+	if (-decpt  > (int)precision) {
+		s = "0";
+		decpt = -precision;
+	}
+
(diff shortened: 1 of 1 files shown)


PaulStoffregen:
Thanks!

PR #219, opened 2017-03-10
Fix malloc corrupting the stack. Added new function reserve_stack() t…

tni:
…o adjust the amount of memory reserved for the stack.

Forum thread:
https://forum.pjrc.com/threads/26559-How-to-detect-when-no-RAM-left-for-a-malloc-It-seems-to-just-keep-going-(Teensy-3-1)

### DIFF
--- a/teensy3/kinetis.h
+++ b/teensy3/kinetis.h
@@ -5794,6 +5794,11 @@ extern void software_isr(void);
 extern void (* _VectorsRam[NVIC_NUM_INTERRUPTS+16])(void);
 extern void (* const _VectorsFlash[NVIC_NUM_INTERRUPTS+16])(void);
 
+// Reserve stack memory. This is the amount of memory malloc will leave alone
+// for the stack. The stack can grow larger than this limit (growing towards /
+// into the heap). 512 bytes are reserved by default.
+extern void reserve_stack(int size);
+
 #ifdef __cplusplus
 }
 #endif
--- a/teensy3/mk20dx128.c
+++ b/teensy3/mk20dx128.c
@@ -32,6 +32,7 @@
 #include "core_pins.h" // testing only
 #include "ser_print.h" // testing only
 
+#include <errno.h>
 
 // Flash Security Setting. On Teensy 3.2, you can lock the MK20 chip to prevent
 // anyone from reading your code.  You CAN still reprogram your Teensy while
@@ -1137,8 +1138,20 @@ void ResetHandler(void)
 
 char *__brkval = (char *)&_ebss;
 
+int reserved_stack_size = 512;
+
+void reserve_stack(int size)
+{
+    reserved_stack_size = size;
+}
+
 void * _sbrk(int incr)
 {
+    if(__brkval + incr >= (char*) &_estack - reserved_stack_size)
+    {
+        errno = ENOMEM;
+        return (void*) -1;
+    }
 	char *prev = __brkval;
 	__brkval += incr;
 	return prev;


PaulStoffregen:
Do this really need to be configurable?

tni:
A lot of people will likely need much more than 512 bytes for the stack (e.g. large buffers declared as local variables in setup()). On the other hand, Teensy LC has very little RAM.

Having to edit mk20dx128.c is rather inconvenient.

PaulStoffregen:
I'm more inclined to go with inconvenient than add another public API to the global namespace, at least for now as a first step on this long-standing problem.

tni:
You could make it semi-public, by just dropping the header file. People could still use their own declaration.

twilkers:
Is there a way to get a linker map file from the build?  While I'm not using malloc(), I wonder in the Ethernet or web server example or other files might be.  

I have a data acquisition application that pushes a web page out over Ethernet.

My application runs for a few hours or a day then stops, wondering if it might be a memory leakage issue like you describe that causes the problem.

Thanks
Tim

PaulStoffregen:
How about this?

    char *__brkval = (char *)&_ebss;

    #if defined(__MKL26Z64__)
    #define STACK_MARGIN  128
    #else
    #define STACK_MARGIN  256
    #endif

    void * _sbrk(int incr)
    {
            char *prev, *stack;

            prev = __brkval;
            if (incr != 0) {
                    __asm__ volatile("mov %0, sp" : "=r" (stack) ::);
                    if (prev + incr >= stack - STACK_MARGIN) {
                            errno = ENOMEM;
                            return (void *)-1;
                    }
                    __brkval = prev + incr;
            }
            return prev;
    }

PaulStoffregen:
Either way, for now I'd prefer to just hard-code the safety margin.  A public API can be added later if it's really needed.  But changing APIs after they've been published can be really painful.  At least initially, I'm really not wishing to add an API for this.

tni:
Your code doesn't prevent memory corruption in cases like:

void fn() {
    char buffer[1024];
    ....
}

malloc(large amounts of memory);
fn();

I would use a much larger stack margin, maybe something like 2kb for T3.2 and 4kb for T3.5/3.6.

PaulStoffregen:
How about these for defaults?

    #if defined(__MKL26Z64__)
    #define STACK_MARGIN  512
    #elif defined(__MK20DX128__)
    #define STACK_MARGIN  1024
    #elif defined(__MK20DX256__)
    #define STACK_MARGIN  4096
    #elif defined(__MK64FX512__) || defined(__MK66FX1M0__)
    #define STACK_MARGIN  8192
    #endif

tni:
Those look reasonable. How about adding a

#ifndef STACK_MARGIN
...

Around your #defines. Or does that count as API?

PaulStoffregen:
Ok, done!  Thanks.

twilkers:
Paul,  

will this be rolled into teensyduino?  If so, when?

Just trying to set my expectation when I should retest

Tim

mikaelpatel:
@PaulStoffregen It would be great if this was a run-time variable instead. This would allow the stack area to dynamically be adjusted if needed. Please see the AVR heap variables and the Arduino-Scheduler for an example. https://github.com/mikaelpatel/Arduino-Scheduler/blob/master/Scheduler.cpp#L94

PaulStoffregen:
Any idea what Arduino is doing on their ARM-based boards, if anything?

mikaelpatel:
I haven't looked into that since I ported the Scheduler to the Due. At the time I couldn't find anything documentation about the heap implementation. Should point out that I didn't really put much effort into that. After reading this issue I started thinking about the setup. I will see if I can find some more info later this week.

pljakobs:
Ah, thanks Paul for including this, it's been a nuisance for a while now.
If I may suggest: could you enable (at least for verbose compiles) "-fstack-usage" for the teensy cores? and possibly provide an override stack frame size #define so we can configure safe smaller or larger stack frames?
the 64k of a Teensy 3.1/3.2 are a lot for a micro, but I still don't like giving 4k of it away just in case if I can actually have better information. 
pj

PR #221, opened 2017-03-16
Adds support for digital pin toggling digitalToggle(uint8_t pin)

awesomenesser:
Based directly off of the existing digitalWrite functions. Adds all defines and functions needed, I have tested it on the Teensy 3.6. It also supports inline digitalToggleFast(uint8_t pin). The hardware has this functionality and I have found it to be useful in my programs.


KurtE:
@PaulStoffregen - Wondering what your thoughts are on this. 

Was just getting ready to debug some ILI9341_t3n code and going to add in some pin toggles and the like, which each time I do something like this, I think about implementing a digitalToggleFast which makes usage of the ports toggle register.  So did a google and saw this PR...

Obviously as this PR is 3 years old, it would need to be expanded on for the newer boards.

PaulStoffregen:
Sure, why not?  3 years ago was a sensitive time, the tail end of the saga of Federico Musto attempting to take over Arduino.  Now I'm much less concerned about extending the API.

KurtE:
Sounds good,  I will look at adding it to the T4 code base

PR #232, opened 2017-04-05
Make Print and Stream constructors constexpr.

tni:
Required for:
https://github.com/PaulStoffregen/Wire/pull/7

### DIFF
--- a/teensy3/HardwareSerial.h
+++ b/teensy3/HardwareSerial.h
@@ -255,6 +255,7 @@ public:
 	virtual void flush(void)        { serial_flush(); }
 	virtual void clear(void)	{ serial_clear(); }
 	virtual int availableForWrite(void) { return serial_write_buffer_free(); }
+	using Print::write;
 	virtual size_t write(uint8_t c) { serial_putchar(c); return 1; }
 	virtual size_t write(unsigned long n)   { return write((uint8_t)n); }
 	virtual size_t write(long n)            { return write((uint8_t)n); }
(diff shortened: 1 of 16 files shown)


PaulStoffregen:
Does this possibly impact Arduino compatibility?

A large number of libraries inherit from these base classes.

tni:
The compiler must support C++11. Arduino has that:
https://github.com/arduino/Arduino/commit/ddf4d878fdb39173b872f69149fb4cf1cd9fd017#diff-0d45b5349ab1c852b7cad3e6b173e880

constexpr constructors can be used for non-const things. The constructor can be invoked with arguments that are not compile-time constants (in which case the resulting object is not initialized during the [constant initialization phase](http://en.cppreference.com/w/cpp/language/constant_initialization)).

The only possible issue is that no new data members that can't be constexpr initialized (with compile time constants) can be added to Print / Stream. There must be a code path where everything is initialized with compile-time constants.

Subclasses won't be affected, except that they now have the option of having constexpr constructors themselves.

tni:
There is a major optimization benefit as well (the unused USB serial and hardware serial code gets eliminated).

**Teensy LC blink, Print / Stream original:**
Sketch uses 12144 bytes (19%) of program storage space. Maximum is 63488 bytes.
Global variables use 2216 bytes (27%) of dynamic memory, leaving 5976 bytes for local variables. Maximum is 8192 bytes.

**Print / Stream made constexpr:**
Sketch uses 6396 bytes (10%) of program storage space. Maximum is 63488 bytes.
Global variables use 2132 bytes (26%) of dynamic memory, leaving 6060 bytes for local variables. Maximum is 8192 bytes.

PaulStoffregen:
I will be traveling for the next 2 weeks.  Please ping me again after May 30th, or in early June.

Before merging this I want to do testing with several libs that inherit from Print & Stream.  The idea looks great, but please understand my cautious nature when it comes to any changes in these base classes, especially changes Arduino hasn't made upstream.

PaulStoffregen:
Ok, going to give this a try...

PR #234, opened 2017-04-06
USB Flightsim for Teensy 4, add FlightSimData and FlightSimEvent data types for T3

jbliesener:
Dear Paul,

I'm one of the more active users of the Teensy flight sim features. I have helped building some fairly complete flight simulators and, along that that path, I've been adding new functions and improving existing ones.

This has led to some improvements of the flight sim plugin and Teensy's core routines that you developed back in 2010(?). We can now handle long dataref names and we also have the (still undocumented) ability to provide user-defined arguments to the onChange function.

In order to be able to support string data stored in X-Plane, I needed to expand the concept of FlightSimInts a little bit.

My first approach (for shorter strings) was to just retrieve up to four string bytes and store it into a FlightSimInt. This required only a single additional line in the plugin code. It also worked some longer strings, but it wasn't very efficient. So I ended up introducing a new read-only data type "FlightsimData" into the plugin and the Teensy core library. This has been working fine for some time and I'd like to release that improvement to the public.

Furthermore, I've added an experimental data type "FlightSimEvent", that I used to interface to MS Flight Sim/P3D. It worked, but it was just a proof of concept (at least until someone wants to pay me to develop it further).

Now, I realize that the Flightsim stuff is not your highest priority. Actually, you mentioned some time ago that you were thinking of abandoning it at all. Here's where my suggestion comes in: I'd like to volunteer as a maintainer for the flight sim part of the Teensy core library (T2 and T3), as well as for the complete plugin. I don't know what's your policy about external maintainers for your core library, but, maybe, we can work something out.

These are the changes for the core library. I'll raise another pull request for the plugin repository. 
Apart from that, I'd be able and willing to take even more responsibility...

Please let me know if you're interested and how we can publish the improvements to a broader audience.

Jorg


jbliesener:
see https://github.com/PaulStoffregen/X-Plane_Plugin/pull/7

jbliesener:
And here we have USB Flightsim for Teensy 4. Implements the original protocol (no change required in plugin). However, as mentioned above, for reading the long datarefs, the new version of the plugin from my GitHub repository must be installed. You're already pointing to it on your web site.

PR #235, opened 2017-04-08
Make the cores main weak.

XMrVertigoX:
Add the weak attribute to the basic main() within cores so others can use that as a subrepo and create their own main method without changing the repo.

### DIFF
--- a/teensy3/main.cpp
+++ b/teensy3/main.cpp
@@ -1,6 +1,6 @@
 #include "WProgram.h"
 
-extern "C" int main(void)
+extern "C" int __attribute__((weak)) main(void)
 {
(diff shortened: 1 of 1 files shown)


PaulStoffregen:
Why is this needed?  The core is compiled to core.a, where any main() function in other code can override it.

Weak symbols are a long-term maintenance issue.  I'm reluctant to use them unless there is a very compelling need.

yoonghm:
See this thread
http://forum.arduino.cc/index.php?topic=43127.0
I tried my own main, it works!!

XMrVertigoX:
The idea was to use this repo as a submodule and build sources directly from here. If the main function is weak, you can add the source files via wildcard.

PR #237, opened 2017-04-15
add SMC_STOPCTRL

FrankBoesing:
https://forum.pjrc.com/threads/43362-SMC_VLLSCTRL-register-is-not-found-in-K66-(Teensy-3-6)


PaulStoffregen:
thanks

PR #238, opened 2017-04-17
Field bits for external wake up in LLWU

yoonghm:


PaulStoffregen:
These are not following the convention where the setting has the register name and field.

They should probably looks like: LLWU_PE_WUPE_PIN_DISABLE

PR #242, opened 2017-05-05
Transfercount plus TNI's aligned attribute

KurtE:
I inclded TNI's aligned attribute fix as it is needed for DMA to work at
all with T-LC...

The new fix with this delta, is there is a problem with the
transferCount method of the DMABase class for T3.x boards.

In particular it checks to see if the count < 512 and does somethig
different which can screw up.   It assmes that if the count is under 512
you are using the ELINK version of the output, which is fine as long as
your previous calls were also under 512.

But with current code, if You call this function in an order like:
Count 6  - Biter and Citer will be set to 6
Count 512 - Biter and Citer will be set to 512
Count 6 - Biter and Citer will be set to 518

Fix is to instead of check for >= 512 to instead check for the ELINK bit
being set then use the 15 bit version else the 7 bit.

Note: potentially should check in the ELINK case that value is < 512,
else it won't work... But not sure what to do in error case?  Currently
if you call with > 32767 it just returns without doing somehing,

But again function does not return any error status, so app will also
likely hang or at best not transfer...

### DIFF
--- a/teensy3/DMAChannel.h
+++ b/teensy3/DMAChannel.h
@@ -41,7 +41,7 @@ extern uint16_t dma_channel_allocated_mask;
 
 class DMABaseClass {
 public:
-	typedef struct __attribute__((packed)) {
+	typedef struct __attribute__((packed, aligned(4))) {
 		volatile const void * volatile SADDR;
 		int16_t SOFF;
 		union { uint16_t ATTR;
@@ -312,7 +312,8 @@ public:
 	// Set the number of transfers (number of triggers until complete)
 	void transferCount(unsigned int len) {
 		if (len > 32767) return;
-		if (len >= 512) {
+//		if (len >= 512) {
+		if (!(TCD->BITER & DMA_TCD_BITER_ELINK)) {
 			TCD->BITER = len;
 			TCD->CITER = len;
 		} else {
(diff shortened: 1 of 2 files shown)


KurtE:
Update: I also update the kinetis.h file for the DMAMUX_SOURCE for the Teensy 3.5 for SPI1 and SPI2

There is only one DMAMUX for each of these and it can be used for either RX or TX, but not at same time.

crteensy:
I second the T3.5 DMAMUX_SOURCE part of this (haven't investigated the rest).

PaulStoffregen:
How's this look?

https://github.com/PaulStoffregen/cores/commit/c697a4456789869a43d557b74780de2e7b2bfe56

KurtE:
Thanks, 

That is complete and should work.

 

The interesting thing is when I was looking over this code (not the fix), it made me wonder things, like how much of this error checking and completeness makes sense?  That is suppose you don’t check for 32767 or 511?  Most likely the calling code will screw up somehow, but with the checks most likely the calling code will screw up somehow, as there is no way for the caller to know it failed… Could change void to bool and return false, but no one will likely look at it… Which is why I asked the question in the thread: https://forum.pjrc.com/threads/44746-Stratagies-for-code-error-detection-and-checking-and-recovery

 

FYI – in my version of SPI code for async support I hacked it like:

#ifdef KINETISK

#ifndef TRANSFER_COUNT_FIXED

inline void DMAChanneltransferCount(DMAChannel * dmac, unsigned int len) {

                // note does no validation of length...

                DMABaseClass::TCD_t *tcd = dmac->TCD;

                if (!(tcd->BITER & DMA_TCD_BITER_ELINK)) {

                                tcd->BITER = len & 0x7fff;

                } else {

                                tcd->BITER = (tcd->BITER & 0xFE00) | (len & 0x1ff);

                }

                tcd->CITER = tcd->BITER; 

}

#else 

inline void DMAChanneltransferCount(DMAChannel * dmac, unsigned int len) {

                dmac->transferCount(len);

}

#endif

#endif

 

Thanks again

PaulStoffregen:
Ok, I'm going to close this one.

PR #245, opened 2017-06-06
analogWrite() can fail with LTO. Fix casts with undefined behavior.

tni:
analogWrite reported hanging with LTO:

https://forum.pjrc.com/threads/44611-analogRead-and-analogWrite-in-one-loop-Teensy-3-6


PaulStoffregen:
Thanks!

PR #250, opened 2017-08-08
end() never disables interrupts

FrankBoesing:
attention, patch for KINETISK only


FrankBoesing:
not disabling interrups means they stay activated and fire even after end() -  a problem with very fast timers.

PR #252, opened 2017-08-15
Event Responder - fix _ms when adding items

KurtE:
When the run time issue adds an item back in to the list, it adds the item at the appropriate time, however it does not decrease the time of the item after where it was to take in account how much time the new item will take before we that next item starts to process it's _ms time

### DIFF
--- a/teensy3/EventResponder.cpp
+++ b/teensy3/EventResponder.cpp
@@ -210,6 +210,7 @@ void MillisTimer::addToWaitingList()
 	bool irq = disableTimerInterrupt();
 	_next = listWaiting;
 	listWaiting = this; // TODO: use STREX to avoid interrupt disable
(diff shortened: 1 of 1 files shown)


KurtE:
This branch now has a couple other fixes. 

If you call timer.begin(500); timer.begin(600);
without a timer tick in between, it would create circular list.  Needed to set state to in waiting queue. 

Also problem if you called begin again when it was in the active queue, when end was called it removed the item from the list, but that messed up timings of items after this item as the timings are cumulative.  So added the item being removed _ms to the item following it in the list. 

Also if you asked for begin(500);  It would go for 502 ticks.  That is because it took one tick to get from waiting queue to active queue, and then you count down to zero and then go one more tick before you actually invoke it.

PaulStoffregen:
Thanks!

PR #253, opened 2017-09-08
Update USB_MTPDISK and white spaces

yoonghm:


yoonghm:
This is for USB_MTPDISK.  Another repo (https://github.com/yoonghm/MTP) is used to test MTP.

PR #274, opened 2018-04-04
Saturate instruction in dac

shkshk90:
Use ARM's unsigned saturate instruction to limit the value, to be written to DAC, to 11 bits (range 0 to 4095).


PaulStoffregen:
Had to revert one of these.  Causes all programs to break on Teensy LC.

https://github.com/PaulStoffregen/cores/commit/27cdeb1cb9d2a26a44fa9cf7169c97944cf39a19

PR #280, opened 2018-07-01
USB user includes

CurrentResident:
This change allows a client project to implement its own `usb_user_desc.h` file and command the USB stack to include it by defining `INCLUDE_USB_USER_DESC` at compile time.

The goal is to simplify overall project management for those that wish to define custom USB interfaces.  With this change, projects have the option of defining custom USB interfaces without having to modify any files that live in the Teensy cores library.

A template header file is included with some instructions to help folks get started with it.  I'm not terribly picky about the wording here - let me know if it needs to be wordsmithed or anything.

Thanks!

### DIFF
--- a/teensy3/usb_desc.h
+++ b/teensy3/usb_desc.h
@@ -115,6 +115,9 @@ If these instructions are missing steps or could be improved, please
 let me know?  http://forum.pjrc.com/forums/4-Suggestions-amp-Bug-Reports
 */
 
+#if defined(INCLUDE_USB_USER_DESC)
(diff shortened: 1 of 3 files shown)


schulzch:
Sounds like a good idea. Two questions:

1. why not the `#include <usb_names.h>` approach?
2. can you descibe compile time definition more precisely? I guess you are referring to CFLAGS `-DINCLUDE_USB_USER_DESC` which usually ends in a messing with Arduinos `.txt` files.

PR #283, opened 2018-09-09
Work-around for C++ tools that do not support type_traits.

wingunder:
The following C++ cross-compiler does not seem to support type_traits
on Debian systems:
   arm-none-eabi-g++ (15:6.3.1+svn253039-1+b1) 6.3.1 20170620

This patch contains a trivial work-around that gets enabled when the
NO_TYPE_TRAITS environment variable is set (to anything) before
running make.


PaulStoffregen:
Is there really any modern toolchain not providing this header?  My understanding is it's supposed to be a required part of C++11 support.

wingunder:
You're right, version 6.3.1 _does_ support type_traits. 
Sorry for this, it turned out to be a toolchain problem indeed.
Closing.

PR #285, opened 2018-09-22
adding bypass functionality to audiostream, to be later used in effec…

trailhead:
Please consider this PR that allows effects derived from AudioStream to be bypassed, and effectively disabled at runtime.  This enables a very rich effects structure, where effects can be added or removed at runtime.  I've tested this with AudioEffectReverb, and will submit a corresponding PR for your reference.  Ultimately I'd like to see this applied to all effects.  In a nutshell, effects will do this...

```
void
AudioEffectReverb::update(void)
{
  audio_block_t *block;

  if (!(block = receiveWritable()))
    return;

  if (!block->data)
    return;

  if (!bypass) {
    // Alter data block only when active.
  }

  transmit(block, 0);
  release(block);
}
```

### DIFF
--- a/teensy3/AudioStream.h
+++ b/teensy3/AudioStream.h
@@ -128,6 +128,7 @@ public:
 	AudioStream(unsigned char ninput, audio_block_t **iqueue) :
 		num_inputs(ninput), inputQueue(iqueue) {
 			active = false;
+			bypass = false;
 			destination_list = NULL;
 			for (int i=0; i < num_inputs; i++) {
 				inputQueue[i] = NULL;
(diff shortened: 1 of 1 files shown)


PaulStoffregen:
Why put this in AudioStream?  Only certain effects would make use of it.  Should probably go into only those effects.

trailhead:
I see your point, but here's the original idea.  Imagine a music making device where the synth objects are assignable to different channels, and have configurable UI screens with common options such as envelope, reverb, etc...  With the statically instantiated nature of the audio library objects, it makes sense to leverage polymorphism through an array of pointers that can be changed.  Then these static objects can be iterated via the pointers, and their common properties can be managed by a flexible UI.  So I put the common property that I wish to use across all effects into the base class.  Maybe moving forward it would make sense to add an intermediate parent class such as "AudioEffect" that groups common attributes.

I couldn't set the bypass property through a common pointer if it were not defined in the base class.  If I were to put the bypass flag into each effect, I would need conditional, type-specific code for every effect I implement bypass for.  Possible, but kinda yucky.

As an extension of this idea, is there any technical reason AudioConnection inputs and outputs couldn't be changed at runtime?

Issue #102, opened 2015-12-19
Invalid assembly generated in delayMicroseconds for teensy lc


expipiplus1:
The instruction `sub %0, #1` is not legal on armv6 unless the register is `SP`

`subs` has to be used instead it seems: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0432c/CHDCICDF.html

PaulStoffregen:
Please provide a program which I can copy into Arduino and compile to see & reproduce the error.

PaulStoffregen:
Maybe you're using a different toolchain?

I can't reproduce this problem when building in Arduino/Teensyduino.

expipiplus1:
Thanks! Sorry for never responding :/

PR #288, opened 2018-10-23
 r0-r7 only for thumb-1 (Bugfix)

FrankBoesing:
Fixes https://forum.pjrc.com/threads/53976-LTO-fails-(Arduino-1-8-5-amp-1-8-7-Teensyduino-1-44)-with-certain-libraries
- might fix https://github.com/PaulStoffregen/cores/issues/102 (wrong description?)


PaulStoffregen:
manually merged due to whitespace

PR #295, opened 2018-12-30
Cycle Counter no longer running on exit

Defragster:
Restart Cycle counter

This is done in configure_systick()
But the counter reads ZERO after current call to usb_pll_start()

I didn't read why - but empirically tested moving it through the code to work

https://forum.pjrc.com/threads/54711-Teensy-4-0-First-Beta-Test?p=194008&viewfull=1#post194008

### DIFF
--- a/teensy4/startup.c
+++ b/teensy4/startup.c
@@ -232,6 +232,8 @@ void usb_pll_start()
 			CCM_ANALOG_PLL_USB1_SET = CCM_ANALOG_PLL_USB1_EN_USB_CLKS;
 			continue;
 		}
+		ARM_DEMCR |= ARM_DEMCR_TRCENA;
+  		ARM_DWT_CTRL |= ARM_DWT_CTRL_CYCCNTENA; // turn on cycle counter
 		return; // everything is as it should be  :-)
 	}
 }


Defragster:
Without a fix for this delayMicroseconds() does not work:

static inline void delayMicroseconds(uint32_t usec)
{
	uint32_t begin = ARM_DWT_CYCCNT;
	uint32_t cycles = F_CPU_ACTUAL / 1000000 * usec;
	// TODO: check if cycles is large, do a wait with yield calls until it's smaller
	while (ARM_DWT_CYCCNT - begin < cycles) ; // wait
}

PaulStoffregen:
Before merging this, I want to dig into *why* the cycle counter is stopping.  It shouldn't, at least as far as I understand.

So please be patient on this one.  I'm going to look into this more deeply before merging this workaround.

Defragster:
Did a quick test on a version of Blink that confirmed the wrap in 7.15x seconds it is not altered (no reset of the cycle count) by calling it once per loop ( each 2 seconds ) or 100,000 times each loop.

Also I removed the counter start from before the return per the pull request and put it both of these places (individually and together as shown) and it leaves a dead T4 because delayMicroseconds() never exits with a stopped cycle counter.:
void usb_pll_start()
{
		ARM_DEMCR |= ARM_DEMCR_TRCENA;
  		ARM_DWT_CTRL |= ARM_DWT_CTRL_CYCCNTENA; // turn on cycle counter
	while (1) {
		ARM_DEMCR |= ARM_DEMCR_TRCENA;
  		ARM_DWT_CTRL |= ARM_DWT_CTRL_CYCCNTENA; // turn on cycle counter

If you want to check further based on the debug prints it is this 'enable USB clocks' case that is taken that stops the Cycle Counter:
CCM_ANALOG_PLL_USB1=80003000
  enable USB clocks
CCM_ANALOG_PLL_USB1=80003040

Though putting it there doesn't work it works well like this at loop top but moved down two lines- but will be executed multiple times:
	while (1) {
		uint32_t n = CCM_ANALOG_PLL_USB1; // pg 759
		printf("CCM_ANALOG_PLL_USB1=%08lX\n", n);
		ARM_DEMCR |= ARM_DEMCR_TRCENA;
  		ARM_DWT_CTRL |= ARM_DWT_CTRL_CYCCNTENA; // turn on cycle counter

In fact if before the 		printf("CCM_ANALOG_PLL_USB1=%08lX\n", n);
it never exits setup.

Defragster:
To confirm - when it dies in my setup - that is with misplaced or NO placement of the CYCCNTENA as follows.
***********IMXRT Startup**********
test 1 -1234567 3
CCM_ANALOG_PLL_USB1=80003000
  enable USB clocks
CCM_ANALOG_PLL_USB1=80003040
Increasing voltage to 1250 mV
need to switch to alternate clock during reconfigure of ARM PLL
USB PLL is running, so we can use 120 MHz
Freq: 12 MHz * 100 / 2 / 1
ARM PLL=80002042
ARM PLL needs reconfigure
ARM PLL=80002064
New Frequency: ARM=600000000, IPG=150000000
ARM PLL = 1200 MHz
ARM divisor = 2
AHB divisor = 1
IPG divisor = 4
USB reset took 5 loops
analogInit
before setup
   F_CPU_ACTUAL=600000000
   millis=145
usb_cdc_line_coding, baud=0
   millis=4000


Where this code is setup:
void setup() {
  pinMode(LED_BUILTIN, OUTPUT);
  //ARM_DEMCR |= ARM_DEMCR_TRCENA;
  //ARM_DWT_CTRL |= ARM_DWT_CTRL_CYCCNTENA; // turn on cycle counter
  Serial4.print( "   F_CPU_ACTUAL=" );
  Serial4.println( F_CPU_ACTUAL );
  Serial4.print( "   millis=" );
  Serial4.println( millis() );
  while ( !Serial && millis() < 4000 ) ;
  Serial4.print( "   millis=" );
  Serial4.println( millis() );
  Serial.println("\n" __FILE__ " " __DATE__ " " __TIME__);
  wCCnt = ARM_DWT_CYCCNT;
}

If I restore the Pull request code debug startup goes to this and code runs as expected:
before setup
   F_CPU_ACTUAL=600000000
   millis=145
usb_cdc_line_coding, baud=0
   millis=4000
after setup

Defragster:
finding that interesting in that it is exiting for timing out :: while ( !Serial && millis() < 4000 ) ;

Not because Serial==TRUE

If I shorten that time (<~500 ms?) tons of these appear: status = 003B0080

If the wait is long enough they clear or never show and it proceeds, if not the Serial.print causes it to never exit setup() - but this is NOT from the CycCntENA

EXCEPT that with no CYCCNTENA 'status = 003B0080' NEVER appears! It just sits.
USB reset took 5 loops
analogInit
before setup
   F_CPU_ACTUAL=600000000
   millis=145
   millis=146
usb_cdc_line_coding, baud=0

Versus with CYCCNTENA:
USB reset took 5 loops
analogInit
before setup
   F_CPU_ACTUAL=600000000
   millis=145
   millis=146
status = 003B0080
status = 003B0080
...

Defragster:
If this is the source of "status =" delayMicroseconds(5); explains the dead stop - it will not exit with stalled CycCnt:
int usb_serial_write(const void *buffer, uint32_t size)
{
	// TODO: do something so much better that this quick hack....
	if (size > sizeof(txbuffer)) size = sizeof(txbuffer);
	int count=0;
	digitalWriteFast(13, HIGH);
	while (1) {
		uint32_t status = (volatile)(transfer.status);
		if (count > 10) printf("status = %x\n", status);
		if (!(status & 0x80)) break;
		count++;
		//if (count > 50) break; // TODO: proper timout?
		// TODO: check for USB offline
		delayMicroseconds(5); // polling too quickly seem to block DMA - maybe DTCM issue?
	}

Defragster:
This is really a bug somehow - just went to older blink without my setup() redoing enable and it just hangs.

Looked in the SDK - source {they use a variant for timing example} and tried dOxygen but search is zilch - and never found any notes in any RM PDF's about the Cycle Counter - or hits on web search of key words.

So I decided to print the values of ARM_DWT_CTRL and ARM_DEMCR before and after re-doing.
The first prints the same value for ARM_DWT_CTRL before and after - so that suggests it is a NOP

However the second ARM_DEMCR goes from 0 to 16777216 - correctly (1<<24)
So only this statement is needed to re-activate the CycleCounter
      ARM_DEMCR |= ARM_DEMCR_TRCENA;

The assumtion (hand waving) I'm making is that when the clock changes to the PLL that changes what feeds the cycle counter so that register is cleared in the process if only for a safe transition as the clock sputters to stability, which may mean it have to delay for the time it takes to get the Lock in PLL.

Given the common usage - even in USB code - of delayMicroseconds() now requiring this to function the results of any failed test in sketch or libray is suspect when the MCU locks.

PaulStoffregen:
I uploaded 1.46-beta4 on the forum, with a fix for this problem.  Please let me know how it works for you?

Defragster:
downloaded now ….

Defragster:
Looks Great ... to the forum

PR #299, opened 2018-12-31
Correct base address of Tempmonitor

mjs513:
Base address should be #define IMXRT_TEMPMON     (*(IMXRT_REGISTER32_t *)0x400D8180)  not
#define IMXRT_TEMPMON     (*(IMXRT_REGISTER32_t *)0x400F8180), my dumb mistake or fat fingers


PaulStoffregen:
Had to manually merge this, since it conflicted with an earlier pull request from Frank.

Thanks!

PR #304, opened 2019-01-01
Renumbered IRQ_RESERVED1-7, IRQ_SOFTWARE = 62

FrankBoesing:


PaulStoffregen:
Don't use IRQ 62.  An earlier (confidential NDA) version of the manual had 62 defined.

Let's use IRQ 70 as "Software".

FrankBoesing:
done :)

PR #305, opened 2019-01-01
Add AudioStream

FrankBoesing:
#define MAX_AUDIO_MEMORY 229376 // same as __MK66FX1M0__
#define AUDIO_SAMPLE_RATE_EXACT 44100
//TODO:
#define CYCLE_COUNTER_APPROX_PERCENT(n) (0)

compiles OK.

### DIFF
--- a/teensy4/AudioStream.cpp
+++ b/teensy4/AudioStream.cpp
@@ -1,0 +1,327 @@
+/* Teensyduino Core Library
+ * http://www.pjrc.com/teensy/
+ * Copyright (c) 2017 PJRC.COM, LLC.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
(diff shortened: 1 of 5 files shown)


PaulStoffregen:
I want to look at this more carefully before merging.  Sorry, just a day or two...

PaulStoffregen:
Going to start looking at this tomorrow.  I want to spend a little time considering a very different internal structure for the buffers, maybe even putting the data in OCRAM for best DMA and the metadata in DTCM for best speed managing the memory.  Even if it all goes into OCRAM, also need to consider the cache row alignment.  I know waiting is painful, but please give me a couple more days on this.... long-term I believe we'll all be much happier if the audio data maps better to this new chip's memory structure.

FrankBoesing:
take your time. the edits I made are very minor and we will have much more to do with the audio library.

FrankBoesing:
Not sure if it makes sense to cache the audio-data.. or if there are more important things to cache. we have "only" 32kb data-cache and perhaps other things are more important? audio is "slow" (compared to the rest)

PR #306, opened 2019-01-01
fast memcpy/memset

FrankBoesing:
tested, works.


FrankBoesing:
Yes, __ARM_ARCH_7EM__ is defined by the compiler., and can be used everywhere.

Just tested, again:
```
#ifdef    __ARM_ARCH_7EM__
    Serial.println("__ARM_ARCH_7EM__");
#endif 
```

PR #307, opened 2019-01-01
Add delayNanoseconds()

FrankBoesing:
This makes sense now, and would be useful for shorter wait-times. Could be used for PulseIn etc.  where we need defined very short delays.


FrankBoesing:
err i meant shiftout, not pulsein.. sry ;-)

PR #308, opened 2019-01-02
optimized delayNanoseconds

FrankBoesing:
shifting both sides by 16 bit gives pretty good accuracy with fast code:
        movw    r3, #9155
        ldr     r2, .L2
        mul     r0, r3, r0
        lsrs    r0, r0, #1
        umull   r3, r0, r2, r0
        lsrs    r0, r0, #8
.L2:
        .word   144122641

with 984 MHz it allows nsec to be max ~143000  max which is plenty


FrankBoesing:
less MHz ->higher allowed nsec without overflow.

FrankBoesing:
error < 10ns

PR #309, opened 2019-01-02
allow high values for usec

FrankBoesing:
allows max usec to be 512 times higher
OK for F_CPU as low as 2MHz


FrankBoesing:
ups.. does not work due to cycle overflow. sorry. 
closed.

PR #310, opened 2019-01-03
T4 - Hardware Serial - Pass 1

KurtE:
This pass has some more support for using the Hardware Serial objects.

Including setting an ISR for doing both reads and writes,  sets up the FIFO,

Sitll lots more work to go.

Things like: fifo - need to better understand/set watermarks, also RX need to read in more than one byte if more available.

Also support fo r other features, like formats, 9 bit, ...

BUt at least you can do simple reads/writes


KurtE:
As I mentioned lots more work to go.  

I want to move all of the objects into separate files, and work with the yield() code to have it such that the objects are only included if you actually use that SerialX object...

Mainly did the PR here, so you can give it a test to see if it works for you as well.

PR #311, opened 2019-01-03
Update AudioStream.cpp

el-supremo:
Fix two bugs in AudioConnection::disconnect.
See https://forum.pjrc.com/threads/54760-Audio-stops-working-when-repeatedly-creating-disconnecting-AudioConnections


PaulStoffregen:
Thanks!

PR #314, opened 2019-01-04
Final Correction to Tempmon Registes and Added analogreadaveraging

mjs513:
Ok.  Thanks to manitou, who caught an error to tempsense2 register offsets got the tempmon to reset the T4 when the panic temperature is reach.  Have verified this.

I also added coded for the analogreadaveraging function in analog.c.  Just as a test I attached a sharp ir sensor to A0 and ran it at 5 and 32 averages.  There wasn't much difference in the average but there is still a difference - with different input:
at 5 point for avg:
avg = 674.7581
var = 93.27432
stdDev = 9.657863

for 32 points:
avg = 675.0081
var = 40.30075
stdDev = 6.348287


mjs513:
Posted this question but not sure if you saw it.  I wanted to updated the
resolution as well but wasn't sure if I changed the res I need to restart
the or do a recal. That manual can be confusing somethings.

Thanks
Mike

PaulStoffregen:
Haven't looked at the tempmon stuff at all.

But I can conform, the manual can be *very* confusing about many things.

PR #315, opened 2019-01-05
T4 SerialX more support (Transmitter enable, Formats, RX Fifo read in...)

KurtE:
Trying to flesh out more of the SerialX support for the Serial1-8 objects.

Things like TransmiterEnable. - Appears to work
Setting CTS/RTS pin - Code in not tested - Actually CTS has issue.
RX Fifo - Now on ISR read all things in FIFO queue... Sets idle, probably needs more testing on when IDLE is set, and clearing the idle state.

Support for the begin(,FORMAT), where now process the format field (At least part of it.  
So far I have tested 8N1, 8N2, 7E1, 7O1.

Still more work needed on testing this, like TX/RX inversion.

### DIFF
--- a/teensy4/HardwareSerial.cpp
+++ b/teensy4/HardwareSerial.cpp
@@ -91,6 +91,29 @@
 #define CTRL_TX_COMPLETING	(CTRL_ENABLE | LPUART_CTRL_TCIE)
 #define CTRL_TX_INACTIVE	CTRL_ENABLE 
 
+// Copied from T3.x - probably should move to other location.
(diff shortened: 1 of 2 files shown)


PaulStoffregen:
Those OneWire macros might need to be renamed or made only visible to the hardware serial code.

KurtE:
Hi Paul, I thought I only put those macros directly into Hardware.Serial.cpp, so that they should only be Seen there. 

Let me know if you want/need anything changed.

PR #320, opened 2019-01-07
T4 - SPI - Create structure

KurtE:
I created a structure IMXRT_LPSPI_t
like what was done for UARTS.

I hopefully got all of the unused fields in properly to get the registers on the correct addresses.
Note: I did not update the other #define LPSPIn_ values but left them, likewise, I create a new define for each of the addresses.

That way we can double check I have everything properly aligned


KurtE:
@PaulStoffregen  - Warning the PR in SPI you took in yesterday relies on the structure defined in this PR to compile.  I could do another PR in SPI if you prefer to temporarily have duplicate of the structure defined if this one is not in place.

PR #321, opened 2019-01-08
alignment

FrankBoesing:
Gives more free DTCM. 
This is the first iteration - after USB and other things are done we should look at this, again.


PaulStoffregen:
an unanswered question: should the vectors be in DTCM or ITCM?  My guess is ITCM might save a few cycles on interrupt entry... but have not tested

PR #323, opened 2019-01-09
Enabled temp monitor in startup

mjs513:
Boiled down the whole temp monitoring sketch to 2 functions,  tempmon_init() and tempmonGetTemp().  The temp monitor gets turned on in startup and then any sketch can just call tempmonGetTemp() to monitor the core temperature as the sketch is running.

One WARNING though the panic temperature is set to 90C, and if exceeded will reset the chip, its easy enough to change though.  Nothing else is done at this point.

### DIFF
--- a/teensy4/core_pins.h
+++ b/teensy4/core_pins.h
@@ -769,6 +769,9 @@ unsigned long rtc_get(void);
 void rtc_set(unsigned long t);
 void rtc_compensate(int adjust);
 
+void tempmon_init(void);
+float tempmonGetTemp(void);
+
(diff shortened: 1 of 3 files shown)


mjs513:
I put back three additional functions, just in case.  (1) to stop measurements, (2) start up measurements again in case you stopped it and the (3) is just to kill it basically, use this and you have to reinit the tempmon.  Was struggling with this but thought there might be cases where you would want to start and stop the process.

PaulStoffregen:
manually merged to get around git merge conflict

PR #326, opened 2019-01-11
T4 digitalWriteFast non constants

KurtE:
The code did not look at the passed in VAL when not constants

Also add Serial7 and 8 to keywords.txt


PaulStoffregen:
manually merged the keywords

PR #327, opened 2019-01-11
Add NULL func to disable debug

Defragster:
See  
https://forum.pjrc.com/threads/54711-Teensy-4-0-First-Beta-Test?p=195229&viewfull=1#post195229

to disable :: //#define PRINT_DEBUG_STUFF

T:\arduino-1.8.8T4_146\hardware\teensy\avr\cores\teensy4\debug\printf.h

Needs this in the ELSE or ifndef case :: #define printf_debug_init()


Defragster:
:)

PR #332, opened 2019-01-17
HardwareSerial: Add NULLptr checks

FrankBoesing:


FrankBoesing:
eh..no.
better fix needed.

PR #333, opened 2019-01-17
Stream find fix handle more cases

FrankBoesing:
- important for plain find(const char *target, size_t length)
- tested


FrankBoesing:
yes.

PR #338, opened 2019-01-22
Drop CPU speed and TEMP when faulted offline

Defragster:


PaulStoffregen:
going to merge this now... but later I'd like to see this done in a way that doesn't alter F_BUS

PR #347, opened 2019-01-27
speed optimize intervalTimer

FrankBoesing:
Much faster (measured!)

### DIFF
--- a/teensy4/IntervalTimer.cpp
+++ b/teensy4/IntervalTimer.cpp
@@ -31,11 +31,10 @@
 #include "IntervalTimer.h"
 #include "debug/printf.h"
 
-static void dummy_funct(void);
 static void pit_isr(void);
 
 #define NUM_CHANNELS 4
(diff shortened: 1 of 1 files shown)


teensy4/IntervalTimer.cpp
```diff
@@ -95,16 +94,23 @@ void IntervalTimer::end() {
 //FASTRUN
 static void pit_isr()
 {
+#if 0
 	for (int i=0; i < NUM_CHANNELS; i++) {
 		IMXRT_PIT_CHANNEL_t *channel = IMXRT_PIT_CHANNELS + i;
-		if (channel->TFLG) {
+		if (funct_table[0] && channel->TFLG) {
```
luni64: Shouldn't that be `funct_table[i]` ?

FrankBoesing: that's disabled.

luni64: Sorry you are right of course. Looks like I'm getting too used to having not defined sections grayed out... 

I'll try it in a minute

PR #357, opened 2019-02-19
Set pad configs: disable keeper

mjs513:
Paul
This addresses the issue identified in post 1782 and 1789 regarding the issue that keeper should be disabled before starting ADC.  This is the easiest way I could think of to implement it.

Mike


mjs513:
I updated the analogread function to include the recommendation to set PKE in that function as well by @manitou: https://forum.pjrc.com/threads/54711-Teensy-4-0-First-Beta-Test?p=201643&viewfull=1#post201643

mjs513:
Paul
To get my version updated I had to delete my copy and redo it after I re-forked.  So I am going to close this and issue a new one.

PR #358, opened 2019-02-21
According to Frank so LTO works

Defragster:
https://forum.pjrc.com/threads/54711-Teensy-4-0-First-Beta-Test?p=198210&viewfull=1#post198210

Handler survives LTO pruning
Remove Printing when PRINT_DEBUG_STUFF #ifdef off

### DIFF
--- a/teensy4/startup.c
+++ b/teensy4/startup.c
@@ -263,6 +263,8 @@ void reset_PFD()
 //  R1
 //  R0
 // Code from :: https://community.nxp.com/thread/389002
+void HardFault_HandlerC(unsigned int *hardfault_args) __attribute__((used));
+
 __attribute__((naked))
 void unused_interrupt_vector(void)
 {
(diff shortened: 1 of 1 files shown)


Defragster:
Paul, LTO may not be working - at least executing - but without this added change it won't compile:

void HardFault_HandlerC(unsigned int *hardfault_args) __attribute__((used));

Defragster:
Cross ref - LTO compile/upload on T4 required 15s reset as noted here: https://forum.pjrc.com/threads/54711-Teensy-4-0-First-Beta-Test?p=202037&viewfull=1#post202037

Defragster:
Paul with the ASM code embedding the call to the Fault_Handler it is missed when 'weak' by LTO parsing.  This (used) decoration with an added 'prototype' suggested by FrankB prevents that loss.

PR #362, opened 2019-03-02
Code cleanup for compiler warnings etc

dfries:
G++ warms on the following when ARDUINO isn't defined, check first.
#if ARDUINO >= 100
void to C function prototypes
use #include "" when the file is relative, otherwise it could get the wrong file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.d
 *.hex
 *.o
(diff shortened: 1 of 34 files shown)


FrankBoesing:
I'd remove the #if ARDUINO >= 100 completely

PR #364, opened 2019-04-17
update setpad in Analog.c and added debug info in startup

mjs513:
Paul
This addresses the issue identified in post 1782 and 1789 regarding the issue that keeper should be disabled before starting ADC. This is the easiest way I could think of to implement it. It was further updated  to include the recommendation to set PKE in that function as well by @manitou: https://forum.pjrc.com/threads/54711-Teensy-4-0-First-Beta-Test?p=201643&viewfull=1#post201643

I enhanced the debug info that @defragster incorporated to further explain fault codes for CFSR and HFSR. I know it works because I tend to be getting them more than others on the 1052.

Mike

### DIFF
--- a/teensy4/startup.c
+++ b/teensy4/startup.c
@@ -308,7 +308,8 @@ void HardFault_HandlerC(unsigned int *hardfault_args) {
   stacked_psr = ((unsigned int)hardfault_args[7]) ;
   // Configurable Fault Status Register
   // Consists of MMSR, BFSR and UFSR
-  _CFSR = (*((volatile unsigned int *)(0xE000ED28))) ;
+  //(n & ( 1 << k )) >> k
+  _CFSR = (*((volatile unsigned int *)(0xE000ED28))) ;  
   // Hard Fault Status Register
   _HFSR = (*((volatile unsigned int *)(0xE000ED2C))) ;
   // Debug Fault Status Register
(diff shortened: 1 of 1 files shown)


Defragster:
I pulled this over to look at - it has shared edits with the PULL opened to allow LTO to compile.
>>  void HardFault_HandlerC(unsigned int *hardfault_args) __attribute__((used));

I see that imxrt.h has #defines for FSR constants included - for the 6 REGS already included: 

Like: _CFSR = SCB_CFSR;  
  // Hard Fault Status Register
  _HFSR = SCB_HFSR;
  // Debug Fault Status Register
  _DFSR = SCB_DFSR;
  // Auxiliary Fault Status Register
  _AFSR = SCB_AFAR;
  // Read the Fault Address Registers. These may not contain valid values.
  // Check BFARVALID/MMARVALID to see if they are valid values
  // MemManage Fault Address Register
  _MMAR = SCB_MMFAR;
  // Bus Fault Address Register
  _BFAR = SCB_BFAR;

Defragster:
The above edits do compile - but unlike mjs513 I've not been seeing any T4 faults so not looked into this.

In fact since I seem to have broken my T4 Beta Breakout board (for display usage) and gone back to breadboard I don't even have a Debug Serial4 link to the T4 since it is running so well.

Good work mjs513 on breakout of those bits … I've got the manual section I linked you to here on the top of my desk pile - I'll look and confirm the math and perhaps see if any others can be brought out.

PR #369, opened 2019-04-26
add RTC functions

FrankBoesing:
TODO: rtc_compensate


FrankBoesing:
Attention, this needs 
__rtc_localtime in boards.txt

FrankBoesing:
hm, there is something missing - github does not let me edit this PR - have to make a new one.

PR #375, opened 2019-05-02
drop out code when no print_debug_stuff

Defragster:
this is (ab)used in HardFault_HandlerC() not under #ifdef


Defragster:
I see you swapped to proper use of printf

PR #382, opened 2019-06-26
Added the ability to parametrize Teensy 3 core USB init delays

stawiski:
Introduced compiler defines: `TEENSY_INIT_USB_DELAY_BEFORE` and `TEENSY_INIT_USB_DELAY_AFTER`; to be able to change the `delay()` values before and after `usb_init();` in `_init_Teensyduino_internal_()` for those who know what they are doing.

This came up as we were developing an application that critically depended on the platform bootup speed (every milisecond was important), and after some search on why does Teensy start up so slowly, we found aforementioned delays. Understandable that lack of those delays causes problems for some people due to their hardware setup, ability to parametrize the delays for those "who know what they're doing" is beneficial.

### DIFF
--- a/teensy3/pins_teensy.c
+++ b/teensy3/pins_teensy.c
@@ -583,18 +583,30 @@ void _init_Teensyduino_internal_(void)
 	TPM1_SC = FTM_SC_CLKS(1) | FTM_SC_PS(0);
 #endif
 	analog_init();
+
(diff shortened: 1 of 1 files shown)


stawiski:
@PaulStoffregen I'll be happy to add that into Teensy 4.0.

PaulStoffregen:
There no need now to keep the numbers for old versions.

PR #384, opened 2019-07-01
Teensy4: fixed serial buffer head and tail types as in Teensy3.

stawiski:
Discovered with @jacobvartanian .

Same fix as my PR for Teensy3: https://github.com/PaulStoffregen/cores/pull/383

### DIFF
--- a/teensy4/HardwareSerial.h
+++ b/teensy4/HardwareSerial.h
@@ -216,10 +216,10 @@ private:
 	size_t  			rts_low_watermark_ = 0;
 	size_t  			rts_high_watermark_ = 0;
 	volatile uint8_t 	transmitting_ = 0;
-	volatile uint16_t 	tx_buffer_head_ = 0;
-	volatile uint16_t 	tx_buffer_tail_ = 0;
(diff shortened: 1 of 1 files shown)


PaulStoffregen:
Is there really a need for buffers larger than 64K?

stawiski:
@PaulStoffregen Yes. We found the bug while developing a logger device, which needs really large ring buffer. We were declaring over 130KB of RAM for that buffer on Teensy 3.
It all depends on your application, you can surely have a reason to.

PaulStoffregen:
I'm not going to unconditionally change the size.  But I will consider extending the conditional approach used on Teensy 3.x.

stawiski:
I can make a compiler switch for Teensy 4 if you want, it'd check all the sizes of all UART buffers and increase the type to 32 bits if any of the sizes is over 64 KB.

Also, why just not use `uint32_t`? Are you trying to save on RAM?
Teensy 3.6 and 4 are both 32-bit cores, so using a native variable might actually produce less/faster assembly as you don't need to emulate overflows. Can you explain reasoning behind `uint16_t`?

PR #386, opened 2019-08-07
WIP: Teensy 3: Create USB Descriptors so we can emulate a Nintendo Switch Controller

earboxer:
(https://forum.pjrc.com/threads/54415-A-usb-controller-adapter-for-the-Nintendo-switch)

To get this to work, edit `boards.txt`

```
teensy31.menu.usb.ns=Nintendo Switch Controller
teensy31.menu.usb.ns.build.usbtype=USB_NS
teensy31.menu.usb.ns.fake_serial=teensy_gateway
```

(for each of the teensy 3s)

Then Just select "Nintendo Switch Controller" from the USB type menu.

The USB Descriptors are based on https://github.com/shinyquagsire23/Switch-Fightstick/blob/master/Descriptors.c (MIT Licensed). (translated by hand).

Joystick.X(value) takes values from 0 to 255.

This also adds Joystick.RX and Joystick.RY for the right control stick.

I changed the parameters for the hat switch: `Joystick.hat(up, right, down, left)` where all of the values are boolean. Nintendo uses a cross-shaped [D-pad](https://en.wikipedia.org/wiki/D-pad) on the Nintendo Switch Pro Controller, which this is mimicking.

Joystick.button should only accept 16 buttons.

### DIFF
--- a/teensy3/usb_desc.c
+++ b/teensy3/usb_desc.c
@@ -237,7 +237,63 @@ static uint8_t mouse_report_desc[] = {
 #endif
 
 #ifdef JOYSTICK_INTERFACE
-#if JOYSTICK_SIZE == 12
+
+#if JOYSTICK_SIZE == 8
+static uint8_t joystick_report_desc[] = {
+    0x05, 0x01,         // Usage Page (Generic Desktop)
+    0x09, 0x05,         // Usage (Joystick, Game Controls)
+    0xA1, 0x01,         // Collection (Application)
+        // Buttons (2 bytes)
+        0x15, 0x00,         // Logical Minimum (0)
+        0x25, 0x01,         // Logical Maximum (1)
+        0x35, 0x00,         // Physical Minimum (0)
+        0x45, 0x01,         // Physical Maximum (1)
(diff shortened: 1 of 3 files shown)


dogtopus:
Some slight problem though:

1. We don't know whether or not Ninty will sue PJRC for "unauthorized use" of their trademark.
2. The code makes teensy pretending itself to be a device made by Hori, which then makes it non-USB compliant.

Both are non-technical but could affect the merge.

earboxer:
1. So the boards.txt should not say "Nintendo". Do we have to change references in the code to "Switch Pro Controller" and "Switch Controller"?
2. What's the threat of being non-USB compliant?

Teensyduino is aimed at making coding easier for the most cases. I think something like
```c
#define VENDOR_ID		0x0f0d
#define PRODUCT_ID		0x0092
#define PRODUCT_NAME     {'M','y',' ','C','o','o','l',' ','C','o','n','t','r','o','l','l','e','r'}
#define PRODUCT_NAME_LEN 18
```

in your `.ino` file works in teensyduino, does it not? (then the user's code would be non-compliant, rather than teensyduino itself).

PR #387, opened 2019-08-12
Simply delayNano Mod Add delayCycles

Defragster:
Demonstrated in https://forum.pjrc.com/threads/54711-Teensy-4-0-First-Beta-Test?p=212280&viewfull=1#post212280

Does this have possible utility worthy of supporting?


Defragster:
It would need tweaking to work on T_3.x's

PR #391, opened 2019-08-31
Speed 24 MHz - not 24 Hz for Lowest voltage

Defragster:
Just reading code and that lone 24 looks a bit short


Defragster:
Tested 24 MHz USB function with code change and saw minor drop in T4 mA power draw with code change: https://forum.pjrc.com/threads/57444-How-to-change-clock-speed-on-Teensy-4-0?p=214257&viewfull=1#post214257

PR #394, opened 2019-09-24
Fixes a lockup issue of the T4 when running sample sketch

mjs513:
Fixes a lockup issue of the T4.  See post https://forum.pjrc.com/threads/57609-Teensyduino-1-48-Beta-1?p=216441&viewfull=1#post216441.

It the current location where you inserted the PIT timer changes it locks up the T4 when you ran one of the example sketches (Demosauce.ino) for the ILI9341_t3, ILI9341_t3n and the ILI9488 libraries.  By moving the PIT time code to after setArmClock it avoids the lockup.


mjs513:
Know your busy Paul but: 
`asm volatile("nop\n nop\n nop\n nop": : :"memory"); // why oh why?`

what????? This fixes it?

PR #395, opened 2019-09-26
On Analog PWM write below or above Resolution - Output LOW or HIGH

Defragster:
User set res to 8 bit and analogWrite(256,13) turned the LED pin #13 off
https://forum.pjrc.com/threads/57758-Teensy-4-0-and-PWM?p=216901&viewfull=1#post216901

Copied code from Teensy3

### DIFF
--- a/teensy4/pwm.c
+++ b/teensy4/pwm.c
@@ -210,6 +210,16 @@ void analogWrite(uint8_t pin, int val)
 
 	if (pin >= CORE_NUM_DIGITAL) return;
 	//printf("analogWrite, pin %d, val %d\n", pin, val);
+	uint32_t max = 1 << analog_write_res;
+	if (val <= 0) {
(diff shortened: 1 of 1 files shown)


Defragster:
One thing of note as written here? :: the pinMode is placed BEFORE the Write?

It was the other way in the Teensy3 codebase?

PaulStoffregen:
Going to wait on this fix.  I want to try integrating with the timers rather than using pinMode to go back to GPIO mode.  Using GPIO mode cause a minor & long-standing issue where all other analogWrite() has the PWM's double buffered behavior, but the GPIO cases immediately override the current cycle.

Defragster:
I wondered why it wasn't in sync with T_3.x code - makes sense it will add a bit of a timing cusp in behavior.

Though now it does present a more significant cusp going off at full value.

PR #397, opened 2019-10-10
T4 - Can now handle sketches with main() instead of setup()/loop()

KurtE:
As per the Teensyduino 1.48 release thread.

T3.x sketches could compile that had int mai() {} and not void() and setup():

So simply copied over setup.cpp from T3 folder of cores, and then changed setup.c to call main, which simply calls setup and loops calling loop/yield.

Added another endless loop after the call to main() to hang there if the users main returns.

Tested with simple sketch
```
#include <arduino.h>
int main (void) {
  pinMode(13, OUTPUT);
  for(;;) {
    digitalWriteFast(13, HIGH);
    delay(250);
    digitalWriteFast(13, LOW);
    delay(250);
  }
}
```
tried it on T3.6 verified before compile failed and now works.  Also compiled and run a normal app to just make sure it was not totally busted

### DIFF
--- a/teensy4/main.cpp
+++ b/teensy4/main.cpp
@@ -1,0 +1,58 @@
+/* Teensyduino Core Library
+ * http://www.pjrc.com/teensy/
+ * Copyright (c) 2017 PJRC.COM, LLC.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ *
+ * 1. The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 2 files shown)


luni64:
Interesting, tested it, works fine.  You can even reduce the ```int main()``` in core/main.cpp to

```c++
extern "C" int main(void)
{
   setup();
   while (1)
   {
     loop();
     yield();	
   }
}
```

Looks like as soon as you define a ```int main()``` somewhere in user code it "overrides" the one from the core. Almost like it was defined as weak which it isn't. I would expect a linker error when there are two mains in the code but the linker does not choke at all and it works perfectly. 
Strange but useful indeed. Any idea why this works?

luni64:
> I believe it works here as well as for Teensy 3.x as,  The first thing the is loaded is the object file for the sketch.  It then  tries to resolve missing symbols that are remaining by trying to load in the  library files that were previously generated.  And since the object file that corresponds to main.cpp, does not have any of the remaining missing symbols, and as such the linker does not bring it it.

Good point. Tested it by directly linking the core object files (not the lib) which  generates the expected linker error (multiple definition of main). That does of course mean that it is a bit fragile (only works with the core files in a lib) but that is probably an academic issue. 

Thanks for saving me from getting mad.

PR #399, opened 2019-10-24
Drop corrupted UART RX data.

stawiski:
> Treat current data as corrupted if any of the following flags is set: noise detected, framing error, parity error

I can make this a conditional compile through a `#define` if you want to.

### DIFF
--- a/teensy3/serial1.c
+++ b/teensy3/serial1.c
@@ -516,6 +516,24 @@ void uart0_status_isr(void)
 {
 	uint32_t head, tail, n;
 	uint8_t c;
+
+	/*
+	UARTx_S1:
+		To clear a flag, the status register should be read followed by a read or write to D register,
+		depending on the interrupt flag type. Other instructions can be executed between the two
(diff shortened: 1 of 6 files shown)


teensy3/serial1.c
```diff
@@ -516,6 +516,24 @@ void uart0_status_isr(void)
...
+
+		Reading an empty data register to clear one of the flags of
+		the S1 register causes the FIFO pointers to become
+		misaligned. A receive FIFO flush reinitializes the pointers.
+		A better way to prevent this situation is to always leave one
+		byte in FIFO and this byte will be read eventually in
+		clearing the flag bit.
+	*/
+	// Treat current data as corrupted if any of the following flags is set: noise detected, framing error, parity error
+	uint8_t isDataCorrupted = (UART0_S1 & (UART_S1_NF | UART_S1_FE | UART_S1_PF)) ? 1 : 0;
```
ssilverman: The Framing Error flag won't be detected because it's on a different IRQ, the "error IRQ", and this isn't enabled by this UART driver.

PR #402, opened 2019-11-01
Update imxrt.h for encoder structure

mjs513:
Hi Paul
Created a Quad Encoder library for the Teensy 4 but to make life simpler I updated the encoder registers to create encoder structure.  Right now basic functions quad encoder function working but probably still needs some testing.

Thanks
Mike


mjs513:
Just for reference the finished lib with some examples is on my GitHub: https://github.com/mjs513/Teensy-4.x-Quad-Encoder-Library

mjs513:
Paul
Per your direction on deleting reg names, "But do not delete the already defined full register names. Convert them to the new struct.", i went back in and added the register names using the typedef structure for IMXRT_ENC_t.  Hopefully, I did it right :).  But seriously if I knew I wouldn't have deleted them. Sorry.

PaulStoffregen:
Thanks for updating this.  I really should have commented earlier... but been so far behind since T4 release.  Finally getting caught up now and back to software work.

mjs513:
Not a problem Paul.  Know you have been crazy busy running a business and trying to do development work.  Good to see you back to the doing the fun stuff :)

PR #404, opened 2019-11-16
T4 avr registers

KurtE:
@PaulStoffregen,

As there have been a few complaints recently recently up on forum about not having support for AVR registers, here is a first pass, maybe last pass... I did not test Pull up stuff yet... But at least simple test works:
```
void setup()
{
  while (!Serial && millis() < 5000) ; // wait up to 5 seconds.
  Serial.begin(115200);
  Serial.println("Quick test to see if AVR emulation works ag all");
  // Quick and dirty test
  // PORTD Pins 0-7
  // PORTB Pins 8-13
  // PORTC Pins 14-19
  DDRD = 0xff;  // All Outputs;
  DDRB = 0;     // All Inputs
  DDRC = 0;     // All Inputs
}

uint8_t counter = 0;
void loop()
{
  PORTD = counter;
  Serial.printf("Count: %x, B:%x C:%x\n", counter, (uint8_t)PINB, (uint8_t)PINC);
  counter++;
  delay(250);
  if (Serial.available()) {
    while (Serial.read() != -1) ;
    Serial.println("Paused:");
    while (Serial1.read() == -1) ;
    while (Serial.read() != -1) ;
  }
}
I have jumpers from pins 0-7 to pins 8-15 and theoutput looks correct where B: gets the lower 6 bits and C gets the upper 2 bits. 

It did find an error in my code where PINC was returning wrong data.  Turned out bug in T3.x code where we were doing digitalReadFast() of the same pin twice and skipping pin... So fixed in both cases. 
```


KurtE:
Quick note: I did try changing the setup code to enable Pull ups like: 
```
...
  DDRC = 0;     // All Inputs
  PORTC = 0xff; // enables PUs? 
```
And the output for port C, showed the 4 bits that were hooked up to nothing as now HIGH 
```
Count: b9, B:39 C:3e
Count: ba, B:3a C:3e
Count: bb, B:3b C:3e
Count: bc, B:3c C:3e
Count: bd, B:3d C:3e
Count: be, B:3e C:3e
Count: bf, B:3f C:3e
Count: c0, B:0 C:3f
Count: c1, B:1 C:3f
```
So it looks like on the surface that part is reasonably working

PaulStoffregen:
Thanks!

Sorry about the delay on this.... it's been a long week of MacOS for me!  Merging stuff now and will work on USB over the next few days, probably leading up to 1.49-beta2 by next week.

PR #411, opened 2019-12-22
Update imxrt.h

KurtE:
Add T4 ADC_ETC definitions for bits plus structure

Decided to go to the format that is closer to the the manual to
and use names like: chain_1_0, chain_3_2...


PaulStoffregen:
Looks good.  :)

Issue #293, opened 2018-12-13
Invalid string descriptor index


mbenkmann:
https://github.com/PaulStoffregen/cores/blob/44359dcf6b3d11fc59f8c386e518db7f1f2a5ad1/teensy3/usb_desc.c#L606

Unless I'm misunderstanding something this seems to refer to a string descriptor with index 4 which does not exist by default.

pierstitus:
The same on line 1104, and in teensy4/usb_desc.c on lines 624, 1131, 1458 and 1965.

lusb -v reports `(error)`, for example:

    Interface Association:
      bLength                 8
      bDescriptorType        11
      bFirstInterface         0
      bInterfaceCount         2
      bFunctionClass          2 Communications
      bFunctionSubClass       2 Abstract (modem)
      bFunctionProtocol       1 AT-commands (v.25ter)
      iFunction               4 (error)

On linux it just tries again or neglects the error. On a stm32 as usb host the enumeration halts on such errors.

So either change all those 4's to 0's, or create a real string for them.

Thanks for the great work!

pierstitus:
Thanks for fixing, any reason you left the MTP descriptor strings out?
teensy3/usb_desc.c line 624 
teensy4/usb_desc.c lines 1131 and 1965

PaulStoffregen:
Opps, missed those.  Been focusing on the Wire lib.

Fixed.  https://github.com/PaulStoffregen/cores/commit/d4d59c98b7b2461f24f5c6a948910ecd811f8157

PR #414, opened 2020-01-07
Add missing serial string descriptors for composite USB devices

geertu:
When creating a composite USB device containing a serial interface on
Teensy 3.2, "lsusb -v" reports an invalid string descriptor in the
serial interface's interface association descriptor:

    Interface Association:
      bLength                 8
      bDescriptorType        11
      bFirstInterface         0
      bInterfaceCount         2
      bFunctionClass          2 Communications
      bFunctionSubClass       2 Abstract (modem)
      bFunctionProtocol       1 AT-commands (v.25ter)
      iFunction               4 (error)

Indeed, usb_descriptor_list[] does not contain an entry for index 4.
Presumably the iFunction field was copied from usb_serial_hid/usb.c,
which does define the string descriptor.

Fix this by adding the missing string descriptor.
This changes the lsusb output like:

    -      iFunction               4 (error)
    +      iFunction               4 Serial

The same bug is present on Teensy 4.0, hence fix it there as well.

Signed-off-by: Geert Uytterhoeven <geert@linux-m68k.org>

### DIFF
--- a/teensy3/usb_desc.c
+++ b/teensy3/usb_desc.c
@@ -1426,6 +1426,8 @@ extern struct usb_string_descriptor_struct usb_string_product_name
         __attribute__ ((weak, alias("usb_string_product_name_default")));
 extern struct usb_string_descriptor_struct usb_string_serial_number
         __attribute__ ((weak, alias("usb_string_serial_number_default")));
+extern struct usb_string_descriptor_struct usb_string_serial_name
+        __attribute__ ((weak, alias("usb_string_serial_name_default")));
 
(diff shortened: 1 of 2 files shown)


PaulStoffregen:
Oh, opps, this was never supposed to have a string descriptor.  The index should be zero.

geertu:
IC.

So the alternative suggestion from https://github.com/PaulStoffregen/cores/issues/293 should be applied instead?

Thanks!

PaulStoffregen:
Yes.  And sorry, this is a been a very low priority bug for a long time.  Fixing it now.

PR #416, opened 2020-01-07
IntervalTimer 'runningFlags' for knowing when to disable IRQ

ssilverman:
This adds tracking of which PIT timer is running so that the PIT IRQ can be disabled appropriately. (Note that this code isn't concurrent-safe.)


ssilverman:
Note that the commit message isn't exactly the same as the PR description. See if it's acceptable to you.

PR #425, opened 2020-01-28
T4 - IntervalTimer - Checks/uses which clock input

KurtE:
The T4 Interval Timer code was hard coded assuming 24mhz clock.  This did not work if the user changed to use Peripheral clock instead like if their sketch did:

```
  // Set PID and GPT to BUS speed
  CCM_CSCMR1 &= ~CCM_CSCMR1_PERCLK_CLK_SEL;
```

So this change instead checks the value of that register to see which clock source is selected for PIT and GPT and then computes based off of that...

Appears to work fine for simple test case:
```
#include <IntervalTimer.h>
// default: IntervalTImer::begin 0.750000 24 17
IntervalTimer my_timer;
uint32_t count = 0;
uint8_t led_state = 0;
void isr () {
  digitalWriteFast(2, HIGH);
  count++;
  if (count == 1000000) {
    led_state = led_state ? 0 : 1;
    digitalWriteFast(13, led_state);
    count = 0;
  }
  digitalWriteFast(2, LOW);
  asm("dsb");
}

void setup() {
  // Set PID and GPT to BUS speed
  CCM_CSCMR1 &= ~CCM_CSCMR1_PERCLK_CLK_SEL;
  pinMode(13, OUTPUT);
  pinMode(2, OUTPUT);
  while (!Serial && millis() < 4000) ;
  Serial.begin(115200);
  if (!my_timer.begin(&isr, 0.5)) {
    Serial.println("\n*** timer begin failed ***");
    while (1) {
      digitalWrite(13, !digitalRead(13));
      delay(125);
    }
  }
}
void loop() {

}
```
Also appears similar (same) code appears to work in @luni TimerTestTool library.

He also found that going to the other clock appeared to reduce the overhead on system???  Not sure but more details up on the thread:
https://forum.pjrc.com/threads/59222-IntervalTimer-on-Teensy-4-0-versus-3-6?p=227971&viewfull=1#post227971

### DIFF
--- a/teensy4/IntervalTimer.h
+++ b/teensy4/IntervalTimer.h
@@ -33,14 +33,13 @@
 
 #include <stddef.h>
 #include "imxrt.h"
+#include "core_pins.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
 class IntervalTimer {
-private:
-	static const uint32_t MAX_PERIOD = UINT32_MAX / (24000000 / 1000000);
 public:
 	constexpr IntervalTimer() {
 	}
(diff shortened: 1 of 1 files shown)


KurtE:
@PaulStoffregen  - It would be nice to get the IntervalTimer capable of going faster.  If need be, I can probably work on fixing the conflicts.  My guess is it is something simple like I pulled in parts of this when we at least got it as fast as a T3.2...

KurtE:
@PaulStoffregen  - decided to go ahead and close this one out... Would need fixing, which I would be happy to do.  But users now also have the option of using TeensyTimer library which allows them to do this.

PR #428, opened 2020-02-04
T4 - USB Keyboard support LEDS

KurtE:
Migrated the T3 code over to handle the message to set the logical keyboard LEDS

As per forum thread: https://forum.pjrc.com/threads/59372-Teensy-4-0-no-support-for-keyboard_leds

### DIFF
--- a/teensy4/usb.c
+++ b/teensy4/usb.c
@@ -621,9 +621,9 @@ static void endpoint0_complete(void)
 	}
 #endif
 #ifdef KEYBOARD_INTERFACE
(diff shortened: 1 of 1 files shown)


PaulStoffregen:
Why is this calling usb_start_sof_interrupts() ?

KurtE:
Sorry, I pasted in wrong line from what was said that worked...
```
#ifdef KEYBOARD_INTERFACE
	if (setup.word1 == 0x02000921 && setup.word2 == ((1 << 16) | KEYBOARD_INTERFACE)) {
		keyboard_leds = endpoint0_buffer[0];
		endpoint0_transmit(NULL, 0, 0);
	}
#endif
```

PR #430, opened 2020-02-08
fix panic-temp power off

FrankBoesing:
The fix of IOMUXC_GPR_GPR16 disabled the power-off on panic-alarm, this should fix it.


mjs513:
Nice Frank.  Thanks for incorporating it.  Probably simplest thing to do.

FrankBoesing:
Yes - I now see it does not switch off 3v3 - i don't know why.

PR #440, opened 2020-02-23
additional Linkerscripts

FrankBoesing:
finally working :)
Link to flash, only FASTRUN is in ITCM. ITCM is zero size if no is
FASTRUN-code  used.


PaulStoffregen:
I'm not going to merge a zero-ITCM linker script.

FrankBoesing:
No problem. Can you say something about the  reason why do you not like it? Coremark is as fast as before, and one can still use FASTRUN which gives ITCM back for the desired function.

FrankBoesing:
No reason?

PR #447, opened 2020-03-08
prepare for MTP

WMXZ-EU:
change usb_desc.c/h for MTPDISK and MTPDISK_SERIAL for both Teensy3 and Teensy4
MTPDISK_SERIAL need own PID
compared to previous PR usb_mtp.c/h is not bundled in user library


PaulStoffregen:
I've updated the USB descriptors, as a starting point to work on bringing full MTP support into the core library in (hopefully) version 1.52.

https://github.com/PaulStoffregen/cores/commit/9dff33654f0d019b16a73063867f7ca9f4684263

Please take a quick look at the config for endpoint 4.  This transmits events to the host, right?  It was configured to receive.  Unless I've seriously misunderstood something, that doesn't look right.

I have long intended to put serious time into MTP... but only after Teensy 4.1, missing audio stuff, and porting the LED libs.  Sorry it's taken so very long.  Now that major progress as been made on all of those, I will soon work on MTP.  The SdFat-only approach is not how I want to provide this to most users.  I also do not wish to have more than 1 USB type defined until it's near the final form (still TBD).  I'll write more about this on the forum soon.

WMXZ-EU:
Re endpoint 4. I indeed thought to have events also receive from the host. that is Teensy could send event to PC to request an action by PC. However, I never got events to work and removed event support from descriptor list.
Re SD-Fat only, It should be not a big deal to modify Storage.cpp in MTP_t4 to support other file systems. For me, I needed exFAT support, that is not provided by stock SD.

WMXZ-EU:
Any reason why you insist on seremu instead of serial. this is a major issue outside Teensiduino, e.g. interacting with Teensy on headless systems.

PR #450, opened 2020-03-12
MPU: Trap not allowed accesses

FrankBoesing:
- trap unused memory regions
- trap nullptr
- trap stack overflow


PaulStoffregen:
These trap regions should be done last, because the hardware gives priority to the higher number ones.

PaulStoffregen:
Added a TODO comment, for later...

PR #454, opened 2020-03-15
Arduino-Compatibility: digitalWrite controls pullup/pulldown in inpu…

FrankBoesing:
…t mode

### DIFF
--- a/teensy4/digital.c
+++ b/teensy4/digital.c
@@ -99,9 +99,12 @@ void digitalWrite(uint8_t pin, uint8_t val)
 		}
 	} else {
 		// pin is configured for input mode
-		// writing controls pullup resistor
-
-		// TODO....
+		// value controls PULLUP/PULLDOWN resistors
+		if (val) {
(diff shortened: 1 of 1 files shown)


WestfW:
I find it depressing that this legacy AVR behavior is getting propagated (usually at significant code expense in digitalWrite())  The pinMode() modification dates back to 2012; we'd be better off finding and fixing the apps that still use the legacy form :-(

FrankBoesing:
Yup, I don't like that much, too. 
But it is needed for better compatibility, some libs may use it, and marked as "TODO" in the sourcecode. It does not have much impact o speed or size. On the T4 (and this PR is for T4) you will hardly notice it, and the prefered digitalWriteFast() stays unchanged.

PaulStoffregen:
@WestfW - Yes, it is depressing, but I'm merging this because some of that very old code is still out there (largely because David Mellis took years to agree to INPUT_PULLUP mainly because of the underscore character)

PR #457, opened 2020-03-21
usb.c / interrupt - callbacks use ctz()

FrankBoesing:


FrankBoesing:
Needs to be tested if it is faster

PR #460, opened 2020-03-28
fix descriptor for teensy 4 usb audio midi serial

widdly:


PaulStoffregen:
Are you sure this is correct?  CDC serial uses the first 2 interfaces.

widdly:
Yes I'm sure, CDC on 0 and 1, MIDI on 2, Audio on 3.  Same as the usb_desc.h in the Teensy_3 directory.

widdly:
Paul, can you take another look at these changes.  The serial midi audio stuff doesn't work in Teensy 4 and it looks to me like a copy-paste error.

PaulStoffregen:
Oh, yup, I see now.  Merging now.

Sorry, been crazy busy since our 2 employees are at home due to Oregon's shelter-in-place order, so Robin & I are struggling to keep PJRC running.

PR #469, opened 2020-05-02
Add recommended 24MHz crystal setting

FrankBoesing:
Later, we should check if we can reduce the bias, also (For me, it works with minimal bias -but I think should should keep it for now, as it is).

### DIFF
--- a/teensy4/startup.c
+++ b/teensy4/startup.c
@@ -52,6 +52,7 @@ void ResetHandler(void)
 	IOMUXC_GPR_GPR14 = 0x00AA0000;
 	__asm__ volatile("mov sp, %0" : : "r" ((uint32_t)&_estack) : );
 #endif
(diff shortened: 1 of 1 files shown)


FrankBoesing:
Page 1175

PaulStoffregen:
Maybe this should be |= ?  Otherwise it will write zeros to all 31 other bits.  Bits 28-26 are described as "This is not a field that the user would normally need to modify".

FrankBoesing:
This sets bits only:
```
#define PMU_MISC0			(IMXRT_PMU.offset150)
**#define PMU_MISC0_SET			(IMXRT_PMU.offset154)**
#define PMU_MISC0_CLR			(IMXRT_PMU.offset158)
```

PaulStoffregen:
Oh, missed the _SET part.

PR #472, opened 2020-05-17
T4.x AttachCTS and setRX can use XBar pins

KurtE:
As there are very very available pins to be used directly as CTS (and
alternate RX pins) and the IMXRT 1062s processor has the ability to
use XBAR pins for these functions.  As per a question from Paul,
I went and implemented it.

I added a pin to xbar mapping table to Hardware Serial.  In actuality
it could be moved to some place more generic as the table is not
specific to SerialX.  Currently the table looks like:

```
//===================================================================
// Should find a good home for this
// Map IO pin to XBar pin...
//===================================================================
// BUGBUG - find a good home
typedef struct _pin_to_xbar_info{
	const uint8_t 		pin;		// The pin number
	const uint8_t		xbar_in_index; // What XBar
	input index.
	const uint32_t 		mux_val;	// Value to set for mux;
	volatile uint32_t	*select_input_register; // Which
		register controls the selection
	const uint32_t	select_val;	// Value for that selection
} pin_to_xbar_info_t;
```
The table for T4 and 4.1 looks like:
```
const pin_to_xbar_info_t PROGMEM pin_to_xbar_info[] = {
{0,  17, 1, &IOMUXC_XBAR1_IN17_SELECT_INPUT, 0x1},
{1,  16, 1, nullptr, 0},
{2,   6, 3, &IOMUXC_XBAR1_IN06_SELECT_INPUT, 0x0},
{3,   7, 3, &IOMUXC_XBAR1_IN07_SELECT_INPUT, 0x0},
{4,   8, 3, &IOMUXC_XBAR1_IN08_SELECT_INPUT, 0x0},
{5,  17, 3, &IOMUXC_XBAR1_IN17_SELECT_INPUT, 0x0},
{7,  15, 1, nullptr, 0 },
{8,  14, 1, nullptr, 0},
{30, 23, 1, &IOMUXC_XBAR1_IN23_SELECT_INPUT, 0x0},
{31, 22, 1, &IOMUXC_XBAR1_IN22_SELECT_INPUT, 0x0},
{32, 10, 1, nullptr, 0},
{33,  9, 3, &IOMUXC_XBAR1_IN09_SELECT_INPUT, 0x0},
{36, 16, 1, nullptr, 0},
{37, 17, 1, &IOMUXC_XBAR1_IN17_SELECT_INPUT, 0x3},
{42,  7, 3, &IOMUXC_XBAR1_IN07_SELECT_INPUT, 0x1},
{43,  6, 3, &IOMUXC_XBAR1_IN06_SELECT_INPUT, 0x1},
{44,  5, 3, &IOMUXC_XBAR1_IN05_SELECT_INPUT, 0x1},
{45,  4, 3, &IOMUXC_XBAR1_IN04_SELECT_INPUT, 0x1},
{46,  9, 3, &IOMUXC_XBAR1_IN09_SELECT_INPUT, 0x1},
{47,  8, 3, &IOMUXC_XBAR1_IN08_SELECT_INPUT, 0x1}
{34,  7, 3, &IOMUXC_XBAR1_IN07_SELECT_INPUT, 0x1},
{35,  6, 3, &IOMUXC_XBAR1_IN06_SELECT_INPUT, 0x1},
{36,  5, 3, &IOMUXC_XBAR1_IN05_SELECT_INPUT, 0x1},
{37,  4, 3, &IOMUXC_XBAR1_IN04_SELECT_INPUT, 0x1},
{38,  9, 3, &IOMUXC_XBAR1_IN09_SELECT_INPUT, 0x1},
{39,  8, 3, &IOMUXC_XBAR1_IN08_SELECT_INPUT, 0x1}
};
```
The code for both attachCts as well as setRX were updated, to
first look to see if the pin matches one that is specifically
is a CTS pin (or RX), then it uses the LPUart system directly,
if not it runs through this list to see if there is a mapping
to pin to an XBar pin, and then configures the Serial system to
use the specified pin as the trigger pin, plus setup XBar to map,
the Xbar pin to the Trigger pin value defined for that LPuart, and
does the stuff to configure that pin for the functionality.

We should note that each XBar signal(?) may map to multiple IO pins,
and only one of them can be used per sketch as XBar pin, example the
signal 17 maps to pins: 0, 5 and T4.1 37.

Also each LPUart has only one trigger value, so you can only use it
for either Cts or RX (or neither).

### DIFF
--- a/teensy4/HardwareSerial.cpp
+++ b/teensy4/HardwareSerial.cpp
@@ -62,6 +62,10 @@
 
 #define UART_CLOCK 24000000
 
+extern "C" {
+    extern void xbar_connect(unsigned int input, unsigned int output);
+}
+
 #if defined(ARDUINO_TEENSY41)   
 SerialEventCheckingFunctionPointer HardwareSerial::serial_event_handler_checks[8] = {nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr};
 #else
@@ -243,9 +247,9 @@ void HardwareSerial::setRX(uint8_t pin)
 				// new pin - so lets maybe reset the old pin to INPUT? and then set new pin parameters
 				// only change IO pins if done after begin has been called. 
 				if ((hardware->ccm_register & hardware->ccm_value)) {
-				*(portConfigRegister(hardware->rx_pins[rx_pin_index_].pin)) = 5;
+					*(portConfigRegister(hardware->rx_pins[rx_pin_index_].pin)) = 5;
 
-				// now set new pin info.
+					// now set new pin info.
 					*(portControlRegister(hardware->rx_pins[rx_pin_new_index].pin)) =  IOMUXC_PAD_DSE(7) | IOMUXC_PAD_PKE | IOMUXC_PAD_PUE | IOMUXC_PAD_PUS(3) | IOMUXC_PAD_HYS;;
 					*(portConfigRegister(hardware->rx_pins[rx_pin_new_index].pin)) = hardware->rx_pins[rx_pin_new_index].mux_val;
 					if (hardware->rx_pins[rx_pin_new_index].select_input_register) {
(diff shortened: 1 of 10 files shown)


KurtE:
@PaulStoffregen  - I am closing this branch as the code was brought in with the PR #474

PR #474, opened 2020-05-20
T4.x Reduce EventResponder overhead on systic and yield

KurtE:
A couple of minimalist changes.

systick_isr - By default it does nothing with the event responder.
Only if the user calls the attachInterrupt member of EventResponder will it install the version of the ISR that checks the list for any active ISR event Responders.

T4 Yield - try to keep bitfields for things yield should test
Make it such that yield hopefully can test to see there is nothing to do and returns quickly.

There are some interesting limitations, that is that the serialEvent
handling code. Both for USB and hardware Serial can only remove it self
if the default eventHandler is called once, as I don't know anyway to
detect if the default code (weak linkage) is included or if it is a user
version.  So waits until called to remove it self from active list.
Sketches can force this by simply calling the event method right after
calling begin on the serial port.

I did a quick and dirty test program to make sure the 3 different ways of using event handler appear to still work. 
```
#include <SPI.h>
#include <EventResponder.h>
#define CS_PIN 10
volatile bool event_happened = false;

EventResponder event;
static const uint8_t buffer[] = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";

void asyncEventResponder(EventResponderRef event_responder)
{
  digitalWriteFast(CS_PIN, HIGH);
  event_happened = true;
  //Serial.println("Event happened");
}

void setup() {
  pinMode(CS_PIN, OUTPUT);
  digitalWriteFast(CS_PIN, HIGH);
  while (!Serial && millis() < 4000) ;  // wait for Serial port
  Serial.begin(115200);
  SPI.begin();
  Serial.println("SPI Test program");
}

void TimeYieldCalls(const char *sz) {
  yield();
  Serial.print(sz); Serial.flush();
  elapsedMicros em = 0;
  for (uint32_t i = 0; i < 1000; i++) yield();
  uint32_t elapsed = em;
  Serial.print(": ");
  Serial.println(elapsed, DEC);
  Serial.flush();
}

void loop() {
  while (Serial.read() != -1) ; // Make sure queue is empty.
  Serial.println("Press any key to run test");
  while (!Serial.available()) ; // will loop until it receives something
  while (Serial.read() != -1) ; // loop until queue is empty
  Serial.printf("start test yield_active_check_flags %x\n", yield_active_check_flags);
  Serial.printf("  systick ISR: %x\n", (uint32_t) _VectorsRam[15]);
  TimeYieldCalls("Start");

  // First try with immediate call.
  event.attachImmediate(&asyncEventResponder);
  Serial.printf("Test Immediate: %x %x\n", yield_active_check_flags, (uint32_t) _VectorsRam[15]);
  event.clearEvent();
  digitalWriteFast(CS_PIN, LOW);
  SPI.transfer(buffer, NULL, sizeof(buffer), event);
  while (!event_happened) ;
  TimeYieldCalls("After Immediate");

  // Use yield .
  event.detach();
  event.attach(&asyncEventResponder);
  Serial.printf("Test Immediate: %x %x\n", yield_active_check_flags, (uint32_t) _VectorsRam[15]);
  event.clearEvent();
  digitalWriteFast(CS_PIN, LOW);
  SPI.transfer(buffer, NULL, sizeof(buffer), event);
  while (!event_happened) ;
  TimeYieldCalls("After yield");

  // Use Interrupt .
  event.detach();
  event.attachInterrupt(&asyncEventResponder);
  Serial.printf("Test Interrupt: %x %x\n", yield_active_check_flags, (uint32_t) _VectorsRam[15]);
  event.clearEvent();
  digitalWriteFast(CS_PIN, LOW);
  SPI.transfer(buffer, NULL, sizeof(buffer), event);
  while (!event_happened) ;
  TimeYieldCalls("After Interrupt");
}
```
Test output shows that yield is slowed down when you install a yield ...
```
SPI Test program
Press any key to run test
start test yield_active_check_flags 1
  systick ISR: 2121
Start: 58
Test Immediate: 1 2121
After Immediate: 57
Test Immediate: 5 2121
After yield: 63
Test Interrupt: 5 2141
After Interrupt: 64
Press any key to run test
```
We can and maybe should do the T3.x as well

### DIFF
--- a/teensy3/EventResponder.cpp
+++ b/teensy3/EventResponder.cpp
@@ -41,6 +41,10 @@ EventResponder * EventResponder::lastInterrupt = nullptr;
 bool EventResponder::runningFromYield = false;
 
 // TODO: interrupt disable/enable needed in many places!!!
+// BUGBUG: See if file name order makes difference?
+uint8_t _serialEvent_default __attribute__((weak)) PROGMEM = 0 ;	
+uint8_t _serialEventUSB1_default __attribute__((weak)) PROGMEM = 0 ;	
+uint8_t _serialEventUSB2_default __attribute__((weak)) PROGMEM = 0 ;	
 
 void EventResponder::triggerEventNotImmediate()
 {
@@ -335,10 +339,17 @@ void MillisTimer::runFromTimer()
 // with libraries using mid-to-high priority interrupts.
 
 extern "C" volatile uint32_t systick_millis_count;
+
 void systick_isr(void)
+{
+	systick_millis_count++;
+}
+
+extern "C" void systick_isr_with_timer_events(void)
 {
 	systick_millis_count++;
 	MillisTimer::runFromTimer();
 }
+
 
 
--- a/teensy3/EventResponder.h
+++ b/teensy3/EventResponder.h
@@ -59,6 +59,7 @@
  * your function is called only one time, based on the last trigger
  * event.
  */
+extern "C" void systick_isr_with_timer_events(void);
 
 class EventResponder;
 typedef EventResponder& EventResponderRef;
@@ -87,6 +88,7 @@ public:
 		detachNoInterrupts();
 		_function = function;
 		_type = EventTypeYield;
+		yield_active_check_flags |= YIELD_CHECK_EVENT_RESPONDER; // user setup a yield type...
 		enableInterrupts(irq);
 	}
 
@@ -112,6 +114,8 @@ public:
 		_function = function;
 		_type = EventTypeInterrupt;
 		SCB_SHPR3 |= 0x00FF0000; // configure PendSV, lowest priority
+		// Make sure we are using the systic ISR that process this
+		_VectorsRam[15] = systick_isr_with_timer_events;
 		enableInterrupts(irq);
 	}
 
@@ -170,6 +174,7 @@ public:
 	EventResponder * waitForEvent(EventResponder *list, int listsize, int timeout);
 
 	static void runFromYield() {
+		if (!firstYield) return;  
 		// First, check if yield was called from an interrupt
 		// never call normal handler functions from any interrupt context
 		uint32_t ipsr;
--- a/teensy3/HardwareSerial.cpp
+++ b/teensy3/HardwareSerial.cpp
@@ -1,6 +1,6 @@
 /* Teensyduino Core Library
  * http://www.pjrc.com/teensy/
- * Copyright (c) 2017 PJRC.COM, LLC.
+ * Copyright (c) 2019 PJRC.COM, LLC.
  *
  * Permission is hereby granted, free of charge, to any person obtaining
  * a copy of this software and associated documentation files (the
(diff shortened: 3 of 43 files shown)


KurtE:
@PaulStoffregen  (@Defragster as you were following on the threads) - Note: this PR became a lot more involved then I originally planned. 

My goal was to reduce the recurring concerns about yield not being fast enough ...
Due to event responder stuff plus the Arduino serialEvent... code 

So I  updated Teensy4 code path to bypass a lot of this work, if those capabilities were not used. 
Example don't need to add ISR code to systic if noone called myevent.attachInterrupt()
No need to call in yield if no one has done an attach() that uses yield. 

Also figured out way to know if the user sketch (or libraries) provided their own serialEvent(X) function for USB or the Hardware Serial.  So again that code is bypassed if none of these are used. 

Worked well enough in T4.x code - That thought I should at least try to get it into T3.x...  So did the same.  Also with this the yield code for doing something like: 
if (Serial1.available()) serialEvent1(), was moved out of yield, and moved into HardwareSerial, where now each HardwareSerial object keeps pointer to it's own serialEventX function and then there is list of the those Serial objects who provide their own serialEvent.  And a method was added to do the test and call code.   What this implies for T3.x and T-LC, is now if your sketch does not use or reference a Serial object, that object, along with its buffers will not be included in the sketch. 

Note: with part of this I pulled in the same changes for CTS pin using XBAR pins of #472

Defragster:
As posted  : [https://forum.pjrc.com/threads/60986-T4-x-(maybe-other)-reduce-serialEventX-overhead-on-system-(worth-it-)?p=240669#post240669](url)

Ran this updated CORES on T_4.1 and T_3.6 - on T_4.1 the next effect looks like equivalent of a void yield() was introduced - a fast running loop() spends about no time in yield() when there are no Event() processing items in use.

FrankBoesing:
Yup, follows more or less my suggestions.
Thank you, Kurt!

Defragster:
Interesting approach Kurt found/used - nice that it also resolves the allocation of code and data space for unused Serial#'s as well! That has been a noted goal in the past.  It seems all Serial# Rx and Tx buffers could move to same default 64 bytes now - and when unused would not allocate any buffer space.

PaulStoffregen:
Looks good.  Large, but good for the common cases where this stuff isn't used.

Do you feel this is ready to merge?  Or should I wait?

KurtE:
Hi Paul, I think it is getting pretty close.  The main thin I or someone else needs to do is to try it out building on Linux or MAC as the build process is sort of interesting

On that it relies on using a weak attribute on variable along with a non-weak one in file that has another weak one and it appears to depend on order of linking or the like.

 

That is for doing the USB one.  With the serialEvent default handler in serialEvent.cpp and the class stuff with the weak variable in a file USB… then I was always getting the non weak version.  But if I moved to other file, that is compiled and linked earlier it worked.  

 

But will be interesting to see if a case sensitive file name system will do it different.  So I will start off pulling out my MAC and setup again with latest stuff and see if it works.

Defragster:
Kurt let me know if you see any way(s) it needs tested for inclusion/merge - with Windows for Serial function.  Compiled okay for 4.1 and 3.6 and only run with no use of Serial# - Event or otherwise.  

With up to 8 Serial's it is a large set of changes and added files.

KurtE:
Fixed the issue I mentioned. Could have fix simply, by removing the default ((week)) serialEventX functions from the HardwareSerialX files, but 

Decide, while at it made the code closer to what I did for the T3.x.  

 

Tried now on both W10 and MAC.  

 

Probably good to go, unless @defragster finds something I missed.

Defragster:
posted a test - looks good and functional - good improvement.

this post #16 and folloing #17: [https://forum.pjrc.com/threads/60986-T4-x-(maybe-other)-reduce-serialEventX-overhead-on-system-(worth-it-)?p=240787&viewfull=1#post240787](url)

Defragster:
Moved test to T_3.6: https://forum.pjrc.com/threads/60986-T4-x-(maybe-other)-reduce-serialEventX-overhead-on-system-(worth-it-)?p=240810&viewfull=1#post240810

Mixed results there - much better wen no events - a bit more overhead when Events in use and slower than TD 1.52

Defragster:
Questions:
T3\yield has checks for dual\triple USB and T4\yield does not? By Design? I don't see it being checked in T4 code?

This  USB_TRIPLE_SERIAL code in T3\yield looks like a typo : line 67?
>		if (Serial**USB1**.available()) serialEvent**USB1**();
should be ?:
>		if (Serial**USB2**.available()) serialEvent**USB2**();

KurtE:
@Defragster  @PaulStoffregen  - Yep - I did not see it in the cores4 branch and so did not add it.

Will do so under new PR unless someone beats me to it.

PR #477, opened 2020-05-24
Fix/Add seralEventUSB1 and serialEventUSB2

KurtE:
The code for T3.x was broken, especially for USB2...

And the code was not setup for T4.x.  So added it.

As @Defragster discovered  and noted in (#474) the USB1 and USB2 code was not in place in the yield code for T4.x so I added it.  

I also found parts of it were missing/wrong for T3.x  so fixed it. 

Had a quick and dirty test sketch that I tested both on with Single, DUAL and Triple serial.  Also where I would rename each of the serial event functions and verify that the event processing flags would properly change. 

```
#include <SPI.h>
void setup() {
  while (!Serial && millis() < 4000) ;  // wait for Serial port
  Serial.begin(115200);
  extern const uint8_t _serialEvent_default;
  extern const uint8_t _serialEventUSB1_default;
  extern const uint8_t _serialEventUSB2_default;
  Serial.printf("Default serialEvent? %d %d %d\n", _serialEvent_default,
                _serialEventUSB1_default, _serialEventUSB2_default);
  Serial.printf("start test yield_active_check_flags %x\n", yield_active_check_flags);

}

void TimeYieldCalls(const char *sz) {
  yield();
  Serial.print(sz); Serial.flush();
  elapsedMicros em = 0;
  for (uint32_t i = 0; i < 1000; i++) yield();
  uint32_t elapsed = em;
  Serial.print(": ");
  Serial.println(elapsed, DEC);
  Serial.flush();
}

void loop() {
}

void serialEvent() {
  Serial.print("Serial: ");
  while (SerialUSB.available())
    Serial.write(Serial.read());
  Serial.println();
}
#if defined(USB_TRIPLE_SERIAL) || defined(USB_DUAL_SERIAL)
void serialEventUSB1() {
  Serial.print("SerialUSB1: ");
  while (SerialUSB1.available())
    Serial.write(SerialUSB1.read());
  Serial.println();
}
#endif
#if defined(USB_TRIPLE_SERIAL)
void serialEventUSB2() {
  Serial.print("SerialUSB2: ");
  while (SerialUSB2.available())
    Serial.write(SerialUSB2.read());
  Serial.println();
}
#endif
```

### DIFF
--- a/teensy3/core_pins.h
+++ b/teensy3/core_pins.h
@@ -2000,9 +2000,11 @@ void _restart_Teensyduino_(void) __attribute__((noreturn));
 // Probably should be in a better spot.  
 extern uint8_t yield_active_check_flags;
 
-#define YIELD_CHECK_USB_SERIAL 		0x1   // check the USB for Serial.available()
+#define YIELD_CHECK_USB_SERIAL 		0x1   	// check the USB for Serial.available()
 #define YIELD_CHECK_HARDWARE_SERIAL 0x2		// check Hardware Serial ports available
 #define YIELD_CHECK_EVENT_RESPONDER	0x4		// User has created eventResponders that use yield
+#define YIELD_CHECK_USB_SERIALUSB1  0x8		// Check for SerialUSB1
+#define YIELD_CHECK_USB_SERIALUSB2  0x10	// Check for SerialUSB2
 
 void yield(void);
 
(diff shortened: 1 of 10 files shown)


Defragster:
That looks better and more inclusive Kurt - just reading diffs again so far ... which let me down before ...

PR #481, opened 2020-07-10
Teensy 4.x compiled Optimize: Debug fails

Defragster:
putting the optimize decoration on configure_cache() allows T_4.x debug compile to run:
See thread https://forum.pjrc.com/threads/61664-No-USB-serial-with-Teensy-4-x-compiled-Optimize-Debug?p=245817&viewfull=1#post245817


Defragster:
**DO NOT DO THIS!** :: This should be an ISSUE not a PR !!!

There is a timing issue - with this in place code that had been working FASTER FAILED

Please see thread above - LAtimes - saw this work on his DEBUG build but it would still fail when nop's were added

PR #496, opened 2020-10-27
imxrt.h - Add CSI Register field names

KurtE:
Started playing with CSI system and got tired of looking at code like:
```
if (last_csistat & (1 << 20)) fb2count++; // increment buffer 2 count
```

Would much rather look at:
```
if (last_csistat & CSI_CSISR_DMA_TSF_DONE_FB2) fb2count++; // increment buffer 2 count
```

So like other subsystems, I went through reference manual and created defines for all of the fields...

Note: I was torn on some of them, as for example: 
#define CSI_CSISR_BASEADDR_CHHANGE_ERROR        ((uint32_t)(1<<28))

Is shown both in the 34.8.7 section to be CHHANGE where probably should be CHANGE

I have not fully converted code over to this yet, so not sure if you want to wait, or assume probably close enough and potentially one or two might need to fixed...


KurtE:
@PaulStoffregen  - With these I assumed we wish to continue with how the register fields are defined through most of the imxrt.h file and not like the SDK does. 

Example: I added: 
```
#define CSI_CSICR1_SWAP16_EN                    ((uint32_t)(1<<31))
```

The 1050SDK (I never downloaded 1060 SDK but assume the same, they have:
```
#define CSI_CSICR1_SWAP16_EN_MASK                (0x80000000U)
#define CSI_CSICR1_SWAP16_EN_SHIFT               (31U)
/*! SWAP16_EN
 *  0b0..Disable swapping
 *  0b1..Enable swapping
 */
#define CSI_CSICR1_SWAP16_EN(x)                  (((uint32_t)(((uint32_t)(x)) << CSI_CSICR1_SWAP16_EN_SHIFT)) & CSI_CSICR1_SWAP16_EN_MASK)
```

PR #500, opened 2020-11-05
T3 x serial x add memory

KurtE:
@PaulStoffregen  - I decided to go ahead and issue a Pull Request to hopefully get this in and tested during this Teensyduino beta phase. 

I added the code to allow T3.x and TLC users to be able to call methods to increase the size of their Serial read and/or write buffers in the same way we added it for T4. 
So it now has calls like Serial1.addMemoryForRead(buffer, sizeof(buffer));

Tested it to see if I broke stuff, by running a Serial daisy chain program where you can hook up TX of one Serial to RX of another and code writes stuff to Serial1 and anything it receives on Serial1 it outputs to Serial...  
Can include code if you want a look. 

The code also allowed you to type in a new baud rate to change all of the Serial ports to.   When I first was trying it, the code would die with new baud rate.  Reduced down to just one Serial and would work for a few baud rates.   

Figured it had to do with the code I added earlier for yield to not directly call the event handlers but instead each Serial object would add itself to the list IF the code was not the default one.  This was done during begin.  Problem was that the second call to begin did not check to see if already in list and the code added again and soon overwrote other memory... So code now checks first to see if already in list.  Also found T4.x had same issue so added that code in there too.

### DIFF
--- a/teensy3/HardwareSerial.cpp
+++ b/teensy3/HardwareSerial.cpp
@@ -39,6 +39,10 @@ uint8_t 		HardwareSerial::s_count_serials_with_serial_events = 0;
 // simple helper function that add us to the list of Serial ports that have
 // their own serialEvent code defined that needs to be called at yield.
 void HardwareSerial::addToSerialEventsList() {
+	for (uint8_t i = 0; i < s_count_serials_with_serial_events; i++) {
+		if (s_serials_with_serial_events[i] == this) return; // already in the list.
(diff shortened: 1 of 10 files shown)


KurtE:
FYI - Test program is in the thread: https://forum.pjrc.com/threads/64325-T3-x-LC-SerialX-Adding-support-for-addMemoryForRead-Write

Issue #184, opened 2016-10-17
Teensy 3.5/3.6 difference found in A10 & A11 pin mapping


soundanalogous:
When working on adding Teensy 3.5 and 3.6 to Firmata Boards.h I found that in pins_arduino.h, [A10 and A11 are mapped to D64 and D65](https://github.com/PaulStoffregen/cores/blob/master/teensy3/pins_arduino.h#L101-L102), but in core_pins.h it appears that [A10 and A11 are mapped to D40 and D41](https://github.com/PaulStoffregen/cores/blob/master/teensy3/core_pins.h#L1270-L1271).

Not sure if this is an issue, but highlighting it just in case.

PaulStoffregen:
Fixed.  Thanks.

PaulStoffregen:
https://github.com/PaulStoffregen/cores/commit/52697d0ef6e44d8d1bc7fbc5e775b3f10f8524c9

PR #504, opened 2020-11-13
Update usb.c so MTP_repsonder can work at 600Mhz

mjs513:
Paul
A long time ago @WMXZ came to the conclusion that the reason why we couldn't see files on the pc side (not sure on LINUX) when using the T4.x above about 450Mhz clock speed was because we had to use the alternate test in usb_transfer_status, see post #184, https://forum.pjrc.com/threads/58033-LittleFS-port-to-Teensy-SPIFlash?p=259193&viewfull=1#post259193.  

@defragster came up with a simpler fix so its only active is you are using mtp:
`file: usb.c
function: usb_transfer_status
changed #if 0 to #ifdef USB_MTPDISK` 

So going to propose the change for the next core update.  Hope i did the pr right :)

### DIFF
--- a/teensy4/usb.c
+++ b/teensy4/usb.c
@@ -999,7 +999,7 @@ void usb_receive(int endpoint_number, transfer_t *transfer)
 
 uint32_t usb_transfer_status(const transfer_t *transfer)
 {
-#if 0
+#if #ifdef USB_MTPDISK
 	uint32_t status, cmd;
 	//int count=0;
 	cmd = USB1_USBCMD;


Defragster:
That looks like the change made here and PC working with LittleFS MTP

PaulStoffregen:
Ok, merging this now - will revisit later...

mjs513:
Thanks Paul - its a workaround for now at least I don't have to keep editing USB.c

PR #505, opened 2020-11-13
Fix dumb mistake on my part - thanksKurt

mjs513:
Fix a very dumb mistake on my part - should have copied and pasted the whole file instead of editing.

Thanks to @KurtE for spotting so fast

Sorry Paul


Defragster:
... oPPs ... I should have caught that ... not certified it as right ...

PR #507, opened 2020-11-17
Update FS.h

WMXZ-EU:
Allow SDFat open modes

### DIFF
--- a/teensy4/FS.h
+++ b/teensy4/FS.h
@@ -165,6 +165,7 @@ class FS
 public:
 	FS() {}
 	virtual File open(const char *filename, uint8_t mode = FILE_READ);
+	virtual File open(const char *filename, oflag_t flags = O_READ);
 	virtual bool exists(const char *filepath);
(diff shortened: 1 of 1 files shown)


PaulStoffregen:
This is exactly the direction I do not wish to take with FS.h.  The whole purpose is to **not** have filesystems specific semantics at this abstract level.

PaulStoffregen:
Just minutes ago I added the missing truncate(size) function.

Let's talk of what the abstract class is lacking which needs this sort of filesystem-specific workaround.  While this sort of thing is "easy", I hope you can appreciate the long-term goal of an abstraction layer which balances the needs of applications needing features against the goal of hiding filesystem-specific details like these open flags.  This is harder to design up-front, but the long-term benefit is people will craft programs which can open files without having to know whether the underlying filesystem is SdFat or LittleFS or a network SMB or others we might support in the distant future.

WMXZ-EU:
You could have replaced oflag_t with uin32_t, then it does not depend on File system.

PR #511, opened 2020-11-21
Teensy models as string defs

FrankBoesing:
Not important, but long standing issue.
Allows to print the Teensy model without adding too much code.
Useful for the MTP-Responder, and perhaps (later) USB-Core to display the correct model.


FrankBoesing:
... not referenced anywhere, but can be done with a simple #include <teensy_model.h>

FrankBoesing:
...would be even easier by using "name"  from boards.txt.  as a -Ddefine in platform.txt

so, closing here (with the hope you add the simple change above to platform.txt)

PR #521, opened 2020-12-14
Allow cores\teensy4 printf debug optionally go to USB

KurtE:
There are times when I wish to debug the cores files, that I wish I could do debug output to USB.  Note: can do with cpp files by doing Serial.printf or the like, but more of a pain with .c files.

The printf stuff put into cores has been very useful for debuging esperailly earlier on.  I understand some of the time it is needed to debug usb stuff or early on stuff so TX pin of Serial4 makes sense.

But for other debugging sessions, wish you had option to go output to the logical Serial object.

So I added a new commented out define in the printf.h file that syas
#define PRINT_DEBUG_USING_USB

That is uncommented, the debug function putchar_debug is defined to either use usb_serial_putchar if we are using normal USB Serial or use usb_seremu_putchar if using Serial emulation.

So far my main testing has been with seremu..

Not sure if this is something you are interested in.  but just in case.
Did this Pull Request

### DIFF
--- a/teensy4/debug/printf.h
+++ b/teensy4/debug/printf.h
@@ -1,4 +1,11 @@
-// #define PRINT_DEBUG_STUFF
+// uncommenting the line below will enable the debug printf statements in cores\teensy4
+// by default it will print to the Serial4 TX pin at baud rate of 115200
+//#define PRINT_DEBUG_STUFF
+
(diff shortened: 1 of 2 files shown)


FrankBoesing:
Very nice!! Would be good to have it. Maybe together with something like this: https://github.com/FrankBoesing/Teensy_Logger

ESP has it, and it is really useful.

PR #524, opened 2020-12-25
Update usb_desc.h

WMXZ-EU:
enable also events to be send from teensy to PC


KurtE:
Hi @WMXZ-EU and @PaulStoffregen 

I have also been playing along with the MTP stuff and have some events working on both T3.x and T4.  
To do this I have also had to update the usb_desc.h files for both teensy3 and teensy4.

My changes are in a branch that also included my changes that allows the cores to optionally do the debug printing to USB instead of Serial4

The actual commit: https://github.com/KurtE/cores/commit/4a17f30e0eb4b3a8b9208a93d533e92e81cec7c4

A couple of differences.  I marked the end point as transmit only (was receive only) as there is only one endpoint defined for this.

Also I increased the event size from 16 to 32.  16 is sufficient for events if only one parameter, but we do have defines for sending events with 2 or 3.  So far have not used them.  Note: my Kindle Fire defined their event size as 28.  

Let me know if I should issue PR of my version or if you might want to incorporate some of these other changes here.

Hope we can get some of the MTP stuff into next beta build.

WMXZ-EU:
Agree to increase MTP_EVENT_SIZE to 32

PR #529, opened 2021-01-03
Fix TEMPMON register offsets in imxrt.h

ssilverman:
The values align with [IMXRT1060RM_rev2.pdf](https://www.pjrc.com/teensy/IMXRT1060RM_rev2.pdf), section 19.4.


ssilverman:
Just force pushed. Rebased onto latest `master`.

PR #530, opened 2021-01-05
AudioStream: Don't disable interrupts globally

FrankBoesing:
Uses NVIC_ENABLE_IRQ / NVIC_DISABLE_IRQ instead, for better interrupt-latency and experience


PaulStoffregen:
Two potential problems.

Failing to disable the higher DMA interrupts during these operations is likely to cause problems.  The DMA interrupts allocate and release blocks

Higher level code which depends on being able to disable the lower priority interrupt across many function calls to the audio library might also fail if these functions unexpectedly re-enable to lower priority interrupt.

FrankBoesing:
Ah, you are right, of course. I did not take the DMA into account. There might be a solution for that. One could e.g. mark the blocks as "no longer needed".
Anyway, my patch was probably a rush job.

PR #533, opened 2021-01-08
Fix alignment error in LPI2C device

hnj2:
There was a small error in the struct representing the LPI2C device registers.
This resulted in reads from the SRDR register to stall the CPU forever.
It probably also prevented any I2C Slave implementation.


PaulStoffregen:
Fixed by https://github.com/PaulStoffregen/cores/commit/654c9270d1820ded46dea2c67105936999030ee5

PR #535, opened 2021-01-11
Print Hardfaults

FrankBoesing:
See forum


FrankBoesing:
https://forum.pjrc.com/threads/64592-Teensyduino-1-54-Beta-5?p=266302&viewfull=1#post266302

Defragster:
Hoping this makes it into TD 1.54 beta 7?

I hacked this on my system and it works to good end!

Was doing 'Wipe All Files/Dirs' in LFSIntegrity with string manipulations and triggered it TWICE in the process with simple dumb code and this restarted and popped up rather than hanging the Teensy.  It was as useful a 'slap in the face' as an error during build. The nature of the failure reported and knowing what code I just added let me quickly resolve it and immediately move forward.

Interestingly working in MTP something was making Teensy restart - but not under this HardFault control? In that case it doesn't help - except to show that there was a code error that was not caused by a trapped fault.

Defragster:
Just remade these edits after Beta 7 install and reviewed the files.

It works here.
`void setup() {
	uint32_t *y = 0;
	y[0]=5;`
Gives:
`Hardfault.
Return Address: 0x26BE
	(DACCVIOL) Data Access Violation
	(MMARVALID) Accessed Address: 0x0 (nullptr)`

Not sure why this is defined if not defined?:
`
// ...\hardware\teensy\avr\cores\teensy4\WProgram.h:
   90  
   91  
   92: #ifndef SHOW_HARDFAULTS
   93: #define SHOW_HARDFAULTS 1`

PR #537, opened 2021-01-28
Add defines for SNVS_LPGPR0-3 and GPR_Z_DIS

KurtE:
There was questions up on forum threads:

https://forum.pjrc.com/threads/66076-Teensy-4-x-s-1062-MCU-has-16-bytes-of-NVRAM-on-RTC-unit?p=268385#post268385

Originally in the thread: https://forum.pjrc.com/threads/64592-Teensyduino-1-54-Beta-5?p=268336&viewfull=1#post268336

That said that the 16 byte memory that was battery backup was not defined nor worked.

So added the defines for the 4 registers.  Only one early on alias was defined before this for the first one...

Also added in the SNVS_LPCR_GPR_Z define which the register bit was talked about in a few other register definitions, this bit was not actually defined in the reference manual, it said it was reserved.  But reads or writes without this bit set in the register will zero out the 4 registers.

Tried example sketch to make sure it worked

### DIFF
--- a/teensy4/imxrt.h
+++ b/teensy4/imxrt.h
@@ -8105,6 +8105,10 @@ typedef struct
 #define SNVS_LPSMCLR			(IMXRT_SNVS.offset060)
 #define SNVS_LPPGDR			(IMXRT_SNVS.offset064)
 #define SNVS_LPGPR			(IMXRT_SNVS.offset068)
+#define SNVS_LPGPR0                     (IMXRT_SNVS.offset100)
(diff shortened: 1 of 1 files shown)


Defragster:
That looks good to match the working code Frank and I posted somewhere in the last year - and that I started that thread with.

The SNVS_LPCR |= (1 << 24) is the magic 'reserved' switch for access you found in SDK and Frank just 'blindly tried' to work.

PR #540, opened 2021-02-12
Audiostream: Waveplayer needs 128Bytes blocks

FrankBoesing:
Prelminary fix until someone fixes the Waveplayer


PaulStoffregen:
Going to wait on this - I want to work with the WAV code soon, might solve this a better way.  Maybe...

FrankBoesing:
Okay. It's just a pity that LC can now use better audio, but the most used feature can't be used because of a small issue!  
But I'll close here.

PR #545, opened 2021-03-09
fix possible round() double promotion

FrankBoesing:
Fixes the case when x is float.


FrankBoesing:
better fix on the way

PR #546, opened 2021-03-16
Fix MPU settings

FrankBoesing:
Two settings with the same adresse, the first with marked readonly, the 2nd as "readwrite". The 2nd has higher priority, so the first is ignored.  
The first is not logical anyway, because it is marked write back and readonly.  
May confuse the MPU?


PaulStoffregen:
This is intentional.  The first 16M is for PSRAM, and then rest is Flash.  It does not confuse the MPU.  ARM designed it so later settings can overlap and override earlier ones.

FrankBoesing:
It seems to fix _something. Reproducable._  
_\- Without: Doom works._  
_\- With: Doom freezes._  
_\- Doom freezes during arm\_dcache\_flush_

_Why "readonly"?_

FrankBoesing:
(freeze happens in an interrupt with priority zero=highest)

FrankBoesing:
Don't we need to set the Subregion bits? (The English here is a bit difficult for me: https://developer.arm.com/documentation/dui0646/a/cortex-m7-peripherals/optional-memory-protection-unit/updating-an-mpu-region?lang=en

FrankBoesing:
I think the MPUs of M4 and M7 are different. 

M4: 8 regions, "overlapping protection regions, with ascending region priority:" https://developer.arm.com/documentation/ddi0439/b/Memory-Protection-Unit/About-the-MPU

M7: 16 regions, "Overlapping regions." [https://developer.arm.com/documentation/dui0646/a/cortex-m7-peripherals/optional-memory-protection-unit?lang=en](https://developer.arm.com/documentation/dui0646/a/cortex-m7-peripherals/optional-memory-protection-unit?lang=en)

(No priority mentioned)

FrankBoesing:
From my undertsand, it would need three entries:

1)

 SCB\_MPU\_RBAR = 0x70000000 | REGION(i++); // FlexSPI2  
SCB\_MPU\_RASR = MEM\_CACHE\_WBWA | READWRITE | NOEXEC | SIZE\_16M;

2)

SCB\_MPU\_RBAR = 0x70000000 | REGION(i++); // FlexSPI2  
SCB\_MPU\_RASR = MEM\_CACHE\_WBWA | READONLY | NOEXEC | SIZE\_128M | (1\<\<8);  
   

3)                         
SCB\_MPU\_RBAR = 0x78000000 | REGION(i++); // FlexSPI2  
SCB\_MPU\_RASR = MEM\_CACHE\_WBWA | READONLY | NOEXEC | SIZE\_128M;  
 

2) sets subregion 1 with 16 MB ( we have 8 subregions qith eqal size. So for 16 MB the main region is 128MB)

3) adds the missing 128 MB

Good night. Late here :)

PaulStoffregen:
Oh, indeed I missed this detail.  Probably best to just use 1 large region and allow writing to all FlexSPI2 memory.

PaulStoffregen:
My initial intention was to make the settings as conservative as possible to (hopefully) fault on programming errors.  Over time, it's always easier to allow more... but later when people have written programs and libraries that do things in bad ways, more restrictive are almost impossible because it will break libraries and hurt people who have already written code.

But indeed I missed this detail.  I'm going to merge this now... and package up 1.54-beta8 by the weekend.

FrankBoesing:
There are two more conflicts. They seem not problematic because there is no cache involved? The nullptr and stack check..

Good night now:)

FrankBoesing:
Most likely we have to remove the first region, too. 

unfortunately, the weblink does not work anymore :( I guess the info was not 100% correct...

PR #547, opened 2021-03-20
Add error message if AudioMemory() is missing

FrankBoesing:
Does it by adding two sections and compare its sizes


Defragster:
Maybe add a commented //audiomemory() from the online GUI in the exported code?

FrankBoesing:
It needs to be in setup()

PR #548, opened 2021-03-20
MPU: Use the correct size specifications

FrankBoesing:


FrankBoesing:
..Still cleaner & better to use this.

PR #550, opened 2021-03-22
make linker values const

FrankBoesing:


FrankBoesing:
sry too early, was a mistake.

PR #554, opened 2021-04-09
Save ram for pin interrupts

FrankBoesing:
Add missing defines and saves RAM by using smaller pin interrupt tables.


FrankBoesing:
Seems not wanted.

Issue #555, opened 2021-04-13
Error in AudioStream::disconnect


mlu:
There seems to be an error in the disconnect method, when unlinking an element from a linked list the pointer of the prevoius node most be shifted, so in AudioStream.cpp line 257

Code:

			if (p == this) {
				if (p->next_dest) {
					p = next_dest;
					break;
				} else {
					p = NULL;
					break;
				}
			}

should probably look like
Code:

			if (p->next_dest == this) {
 				p->next_dest = this->next_dest;
				break;
			}

PR #556, opened 2021-04-14
Makes AudioConnection robust to new, delete, connect and disconnect

h4yn0nnym0u5e:
- Fixes issue #555 . 
- Constructor and destructor code more robust
- Attempts to ensure AudioConnection is completely unlinked from any associated AudioStreams before it is deleted from memory. 
- Corrects an apparent bug where it is possible to specify an input channel one higher than the maximum provided. 
- Permits re-connection of disconnected AudioConnection to a different source and destination to the ones specified at the time it was constructed.
- Returns 0 for successful connect() and disconnect() call, otherwise (undocumented!) error code


h4yn0nnym0u5e:
Just realised this changes a documented feature: "If you attempt to connect many outputs to a single input, only the last connection will work.". Not sure if this is likely to break many applications, and suspect it's not a great feature as presumably the previous connections are now in limbo, thinking they're connected when in fact they're not.

PaulStoffregen:
I have have merged this for 1.56, but I do have some lingering concerns.  Going to merge it now so we can get it into the first 1.57 beta.

My 2 main concerns....

1: The constructor & destructor may not be disabling interrupts enough.  While creating and destroying connections from within interrupts seems a little crazy, I'm pretty sure people will attempt it.

2: The AudioDebug class has the effect of giving public access to everything which should be private.  That's fine for testing, but not something which can be part of a Teensyduino release.

h4yn0nnym0u5e:
> I should have merged this for 1.56, but I do have some lingering concerns. Going to merge it now so we can get it into the first 1.57 beta.
> 
> My 2 main concerns....
> 
> 1: The constructor & destructor may not be disabling interrupts enough. While creating and destroying connections from within interrupts seems a little crazy, I'm pretty sure people will attempt it.
> 
> 2: The AudioDebug class has the effect of giving public access to everything which should be private. That's fine for testing, but not something which can be part of a Teensyduino release.



1. Not a big fan of disabling interrupts, especially as there's no decent agreed method for nesting the disable/enable pairs so only the last enable is effective. Plus I believe the main source of instability is going to be interrupts which occur _during_ construction/destruction, so people doing it in an ISR are probably _safer_! Though still crazy... I've tried to order the code such that an audio update will always find the linkages in a safe state
2. Agree, should at least be #if defined() out. It is massively useful for chasing obscure internal bugs, so worth being able to re-instate easily.
3. I did find a bug relating to NULL src/dst pointers, but I _think_ it only impacts the full Dynamic library where the AudioStream object may no longer exist
4. Note that the changes weren't made for Teensy 3.x at the time of the PR, only 4.x. The code changes are identical, I can make those if you wish

h4yn0nnym0u5e:
Above changes done and PR#642 generated

PR #564, opened 2021-06-07
Fix T4.1 PWM pin number in pwm.c

shirriff:
FLEXPWM3_PWM3_B is assigned to GPIO_EMC_22 (manual p302). This is wired to Teensy 4.1 Flash CS (according to the
schematic) which is pin 51.
The pinout reference card shows PWM on pin 51, not 41, so I think this is a typo.


PaulStoffregen:
Yup, a typo.  Thanks.

PR #566, opened 2021-06-20
Update CrashReport.cpp to print detail info

mjs513:
Hi Paul, @KurtE @Defragster 

As mentioned I tried the unmodified version of crash report with the following:
`void setup() { delay(5000); uint32_t *y = 0; y[0]=5;}`
and it looks like it faulting - getting a blinking LED and nothing being printed in loop (i added some prints in loop just a test), but nothing is being printed to the serial monitor.

Anyways, I went ahead and added additional print info for CFSR and HFSR like Frank and the previous core.

I did not add the tempmon attachinterrupt or the unusedISR faults.  Wasn't sure if you wanted to add that to startup.c.

### DIFF
--- a/teensy4/CrashReport.cpp
+++ b/teensy4/CrashReport.cpp
@@ -13,7 +13,7 @@ struct arm_fault_info_struct {
         uint32_t xpsr;
         uint32_t crc;
 }; */
-
+extern unsigned long _ebss;
 
 size_t CrashReportClass::printTo(Print& p) const
 {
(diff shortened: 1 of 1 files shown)


mjs513:
That was fast.  Ok it now works with instructions from Paul:
Sketch
```
void setup() {
  Serial.println("Testing Crash Report");
  delay(5000); 
  Serial.print(CrashReport);
  uint32_t *y = 0; y[0]=5;
}

void loop() {
  Serial.println("cccc");
  delay(5);
}
```
Output in Serial monitor.
```
CrashReport ... Hello World
  length: 9
  IPSR: 3
  CFSR: 82
	(DACCVIOL) Data Access Violation
	(MMARVALID) Accessed Address: 0x0 (nullptr)
  HTSR: 40000000
	(FORCED) Forced Hard Fault
  MMFAR: 0
  BFAR: 0
  return address: C4
  XPSR: 61000000
  crc: A57C5478

```
Now back to playing.

PR #571, opened 2021-06-21
Update CrashRPT for Temp, Time, Panic temp and a cool off

mjs513:
Hi Paul
I added some more stuff into the crash report tested temp and RTC - so not sure ready for prime time but you can play with it some.  
1.  It pulls the temp at the time of the crash.
2.  Pulls in RTC time of the crash.
3.  Adds an additional note when the reboot cause is by temp (panic temp exceeded).
4.  I take a shot at adding in a cool down before reset but now sure thats going to work.  Have a look

One issue I found was the isvalid is always returning 0 - haven't really looked at why yet. But put in a workaround just so it trips for testing:
`if (info->len > 0) {`

Ok have a few other things to do.

### DIFF
--- a/teensy4/CrashReport.cpp
+++ b/teensy4/CrashReport.cpp
@@ -29,7 +29,6 @@ size_t CrashReportClass::printTo(Print& p) const
 	uint8_t mm = info->time % 60;
 	info->time /= 60;
 	uint8_t hh = info->time % 24;
-	p.println(__TIME__);
 	p.printf( "%02d:%02d:%02d\n", hh, mm, ss  );
 
 	p.print("  Temperature at time of fault: ");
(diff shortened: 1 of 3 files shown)


PaulStoffregen:
Please change "temp" type to float.  Don't use int16_t.  All variables in that struct are supposed to be 32 bits.

mjs513:
Done and done and did a quick check as well.

Don't know how you do this, think I said it before.

PaulStoffregen:
Right now this is the easy part.  Getting chips to put Teensy 4.1 back into production is today's main challenge!

mjs513:
Sorry Paul - just saw the message.  Yeah - compared to getting parts these days, this is the easiest part!  Don't envy you that job.

PR #573, opened 2021-06-23
Shutdown on Panic Alarm after CrashRPT printed.

mjs513:
Evening Paul - @Defragster  - @KurtE 

After a few headaches  and some help from Tim and reading the RM again I got the Crashrpt to trip on the panic alarm and printed before I decided to just do a shutdown instead of a cool off before report.  I posted this question on line as well but figure you guys are hours earlier than me.

Reason - you don't know what caused the over temp - just the overclocking or something more dangerous

### DIFF
--- a/teensy4/CrashReport.cpp
+++ b/teensy4/CrashReport.cpp
@@ -14,6 +14,8 @@ struct arm_fault_info_struct {
         uint32_t crc;
 }; */
 extern unsigned long _ebss;
+extern "C" bool temperature_is_safe(void);
+extern "C" uint32_t set_arm_clock(uint32_t frequency); // clockspeed.c
 
 static int isvalid(const struct arm_fault_info_struct *info);
 
@@ -160,7 +162,13 @@ size_t CrashReportClass::printTo(Print& p) const
   }
   if (SRSR & SRC_SRSR_TEMPSENSE_RST_B) {
     p.println("Reboot was caused by temperature sensor");
-	if(CCM_ANALOG_MISC1_IRQ_TEMPPANIC == 1) p.println("Panic Temp Exceeded");
+	  SRC_SRSR &= ~0x100u; /* Write 0 to clear. */
+	  p.println("Panic Temp Exceeded Shutting Down");
+	  p.println("Can be caused by Overclocking w/o Heatsink or other unknown reason");
+	  IOMUXC_GPR_GPR16 = 0x00000007;
+	  SNVS_LPCR |= SNVS_LPCR_TOP; //Switch off now
+	  asm volatile ("dsb":::"memory");
+	  while (1) asm ("wfi");
   }
   clear();
   return 1;
--- a/teensy4/startup.c
+++ b/teensy4/startup.c
@@ -515,6 +515,7 @@ extern void usb_isr(void);
 //  R0
 // Code from :: https://community.nxp.com/thread/389002
 
+
 __attribute__((naked))
 void unused_interrupt_vector(void)
 {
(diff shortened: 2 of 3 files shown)


Defragster:
@mjs513 
Nice the calling " unused_interrupt_vector(); " was a working path!

I see it stalls with " while (1) asm ("wfi");" when restarts on over temp - shouldn't it drop CPU speed  help cooling?
 > But that only helps when user calls : Serial.print(CrashReport);
 - without that it will just resume ???

I see this went away :  if(CCM_ANALOG_MISC1_IRQ_TEMPPANIC == 1) {
but no delay in 8 sec restart to make sure the temp  gets lowered before exit?

PaulStoffregen:
Looks like we still need checking to only start the 8 seconds when temperature is safe.

PaulStoffregen:
I want to "feature freeze" startup.c later this week, in prep for 1.54.  Plenty can still be done in CrashReport.cpp.  But if we're going to get to a stable 1.54 release soon, we really need to soon cut off more changes to the higher risk code.

Defragster:
Can  p.print  tell when the data actually gets 'stream'ed?

Right now it does the clear() no verification Serial or whatever the output device was receptive.

PaulStoffregen:
No, there is no API to query buffer status.  But the API does have a flush() function which waits (or is supposed to wait) until all data is actually transmitted.

Defragster:
Figured not ... doing a flush() that waits forever would be bad.

> perhaps auto clear is weak - better to see it twice and then have it clear()'d, than to not get to see it at all if the user codes it wrong - they still can if they do a clear() with a ready stream.
Given how easy it is to call Serial.print(CrashReport);  it wasn't much harder to overtly call .clear() when the user so chooses confirming the Report was seen ... as the user sees fit to code it.

Defragster:
The only thing I'm wondering is startup.c might have is in https://forum.pjrc.com/threads/66403-Re-enable-bootloader-interrupts-after-a-hard-fault-Teensy-4-1?p=281851&viewfull=1#post281851

Though of course without the call to CrashReport - then there would be 'fast' repeated reboot. Unless there was a 'weak' call the user could add to sketch to replace the default wait for reaction of their choosing.
> That "weak void userCrashReport() { // on prior Crash wait 8 secs }" then enter setup()
> Sketch "void userCrashReport()" - could wait for Serial - or other - and then stream the output to a place of their choosing - that might solve KurtE's issue where it could go to elsewhere.
Not a simple as "if ( CrashReport ) {" - but that would still work, and pasting a template copy of userCrashReport() could give more control?

mjs513:
@PaulStoffregen and @Defragster 
Between sleep cycles and doing this response - I must be crazy.  In response

> Looks like we still need checking to only start the 8 seconds when temperature is safe.

Right now on panic alarm there is no reboot.  The T4.1 shuts down on panic alarm with the following messages:
>Reboot was caused by temperature sensor
>Panic Temp Exceeded Shutting Down
>Can be caused by Overclocking w/o Heatsink or other unknown reason

If you still want it to restart as opposed to shutdown let me know.  I did mention this up on the thread as well.  Tim seemed to agree:
>Shutdown seems right.

As I said I am leery of continuing to reboot when the cause for the panic temp is unknown.  Before the merge tempmon would shutdown the T4 if the panic temp was reached.   Agree more than likely its due to overclocking but ....

Defragster:
Mike: is this code in the current PJRC github as last merged?

mjs513:
Tim: the shutdown on panic alarm. Yes it us.  For testing I made set him temp to 50 and panic temp to 55 and ran over clicked at >800

Defragster:
Okay - saw both githubs same 4 hrs old.

It is against first plan to restart - shutdown seems safe path. In your test case of a heating lamp dropping speed wouldn't help - in the real world it might be a fan or airflow failure causing over temp.

Only issue is that the crash data won't survive is power shutdown happens for system restart.

But an ill behaving Teensy in any oddly controlling of critical stuff could wake up and half baked and do bad stuff without supervision.

Also not sure what the Teensy state of I/O is in that 8 second time period?  No interrupts or programmatic response can happen - what is the state of the outputs? Are they maintained as last set? Assuming a proper design would fail safe if the Teensy went off/poof ... what is it 

Would be nice to commandeer 11 DWORDS of EEPROM for more permanent recall.

mjs513:
Tim - see my response in https://forum.pjrc.com/threads/66403-Re-enable-bootloader-interrupts-after-a-hard-fault-Teensy-4-1?p=281920&viewfull=1#post281920

PR #576, opened 2021-06-27
move unused int vect to flash

FrankBoesing:
Saves some 100 Bytes RAM


FrankBoesing:
?

PaulStoffregen:
Undecided on this.  One of the failure cases I want to explore is what happens if the FlexSPI config gets messed up.

But this week I am catching up to many important tasks which were set aside to focus on the 1.54 software release.

PR #578, opened 2021-06-28
Allow to implement endpoint number 1

phdussud:


phdussud:
Hello, I need to implement a USB device that is a clone of an existing one. I need to implement some endpoint number 1 for an interface. As the code stands, the first supported custom endpoint is 2.

PaulStoffregen:
Endpoint 1 is reserved for a planned future feature

phdussud:
The boiler plate code (reflected by my changes) would still be valid, right?

phdussud:
But the boiler plate code would still be correct, right?

PaulStoffregen:
When Teensy gets this future feature, this code will not work.  I can not merge it.

phdussud:
I see.. 
Thank you very much for your time. 
Patrick.

PR #582, opened 2021-07-07
(T4) usb_desc: Allow user defines

FrankBoesing:
See https://forum.pjrc.com/threads/67663-Usb-Vendor_id-manufacturer_name?p=283112&posted=1#post283112


FrankBoesing:
Thx to vmuzik, seems that is not needed.

PR #584, opened 2021-07-09
Usb serial fix

FrankBoesing:
Fix serial by blocking autoflush for a short time


FrankBoesing:
sorry.. did help a little bit only.. first liens were ok, but errors occured later.

PR #587, opened 2021-07-19
USBAUDIO: Support other sample freqs

FrankBoesing:
Tested with 44.1k, 44k, 88.2k, 96k, AudioInputUSB

In theory, every freq that is supported by the OS should work.


FrankBoesing:
Minimal changes only, 44.1 is unchanged.

FrankBoesing:
Tested AudioOutputUSB, too, works.

FrankBoesing:
General interest seems pretty low, and poeple have problems. I'll try later again..

PR #593, opened 2021-08-14
Tmod p38 39 set rx

KurtE:
Paul, 

I thought I issued a PR on part of this.  Two parts:

The original MMOD code sort of assumed that the SDIO pins are the same as they are on T4.0.   But turns out that the pins are almost the same.  However pins 38 and 39 are swapped.  I found this out when I rang out all of the pins on my breakout board and these rang out flipped. 

So updated the table for this.  Also as these are Serial pins the TX, RX pins needed to be swapped, in HardwareSerial5.  Likewise PWM pin data needed to be updated.

Also they are XBar pins, so needed to update these pin in the XBar table in HardwareSerial.cpp.  

During testing of this I found that the setRx pins to these was not working.  Turns out there is an issue with this code that can only be called after SerialX.begin.  Also can not call begin again... So fixed those plus updated the table.


KurtE:
@PaulStoffregen  - Would be good to pick this up as well for next beta, as the pin numbers are plain wrong on these pins.

@mjs513 and myself tested through these changes, such that for example a pin test to your breakout board gets the right pin numbers for these two pins.

PR #594, opened 2021-08-17
make systick_safe_read a local var

FrankBoesing:


PaulStoffregen:
I'm pretty sure this change would break the synchronization mechanism.  The intention is for a single systick_safe_read to be shared between main program and all interrupt priority levels.

FrankBoesing:
I'm not, its only used in micros(). It#s value does not matter.

PaulStoffregen:
More than 1 instance of micros() is possible, if used from main program and within interrupts.

FrankBoesing:
It's used as a pure interrupt-detector here. if a higher prio interrupt occurs, the loop repeats. that works with a local var, too... not sure how to write a short demo that shows that..

PaulStoffregen:
Here is the conversation where the change from disabling interrupts to use of LDREX was discussed.

https://forum.pjrc.com/threads/54711-Teensy-4-0-First-Beta-Test?p=195925&viewfull=1#post195925

At the time, you wrote "disabling interrupts works - but... it disables interrupts. not good."

FrankBoesing:
the trick is just, that strex fails when an interrupt happened betwen ldrex and strex.

PaulStoffregen:
Ok, I believe what you're saying is this works even if not used on the same variable.  That probably is true for the M7 core we have now on Teensy 4.0 & 4.1 & MicroMod.  But it may not necessarily be true on future chips.  My understanding of LDREX is it _should_ be used on the same memory location.

FrankBoesing:
You're right, and the 4 bytes are not worth to make too much words :)

PR #595, opened 2021-08-20
Arduino compatibility

FrankBoesing:
From [https://forum.pjrc.com/threads/68011-Error-Compiling-SerialTerminalPRO-example?p=286526&viewfull=1#post286526](https://forum.pjrc.com/threads/68011-Error-Compiling-SerialTerminalPRO-example?p=286526&viewfull=1#post286526)

resp

https://github.com/miko007/SerialTerminal/issues/3#issuecomment-902610304

Makes this work:

```
void setup() {
  String foo = "Hello";
  for (auto& car : foo) {
    Serial.print(car);  
  }
};
 void loop(){};
```

### DIFF
--- a/teensy3/WString.h
+++ b/teensy3/WString.h
@@ -166,6 +166,10 @@ public:
 		if (!buffer) return &zerotermination; // https://forum.pjrc.com/threads/63842
 		return buffer;
 	}
+    char* begin() { return buffer; }
+    char* end() { return buffer + length(); }
+    const char* begin() const { return c_str(); }
+    const char* end() const { return c_str() + length(); }
 
 	// search
 	int indexOf( char ch ) const;
--- a/teensy4/WString.h
+++ b/teensy4/WString.h
@@ -166,7 +166,11 @@ public:
 		if (!buffer) return &zerotermination; // https://forum.pjrc.com/threads/63842
 		return buffer;
 	}
-
+    char* begin() { return buffer; }
+    char* end() { return buffer + length(); }
+    const char* begin() const { return c_str(); }
+    const char* end() const { return c_str() + length(); }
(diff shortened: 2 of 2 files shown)


FrankBoesing:
Confirmed to fix compiliation errors by https://github.com/miko007/SerialTerminal/issues/3#issuecomment-903001368

patrick-dojofive:
👍

PaulStoffregen:
Hi Frank.  Saw your comment on the forum just now.

Hope you'll come back after a good rest.

Indeed a huge number of issues are open.  Also many promised features are still missing.  We pushed very hard to get Teensy 4.0 and 4.1 released and made perhaps too many sacrifices along the way - so many things to do after release.  But then the pandemic happened and just keeping PJRC running without staff drained away all dev hours for over a year!

I hope you can see things are starting to improve.  I know you personally don't care about code security, but it was one of many important features which were dropped in mid-2019, to keep from further delaying Teensy 4.0 release.  Now, near the end of 2021, it's finally happening.

So too will a lot of other long-promised features.

PaulStoffregen:
I'm bringing this into Teensy's core library now (so it'll be in 1.55-beta3), but not quite as implemented in this pull request (which appears to be just copying Arduino's implementation).

If you look at the comments in the code just a few lines up, you'll see the link to this forum thread:

https://forum.pjrc.com/threads/63842

We've been down this road before, where directly return the buffer address (as this pull request does) leads to people dereferencing a null pointer.  This is why I wait to review pull requests, until I have time and I'm not in a hurry.

FrankBoesing:
Thanks for the fix of my fix. No, I always welcome that when you look closely at PRs.

PR #599, opened 2021-08-24
Fix GCC10 -O3 crash in startup.c

FrankBoesing:


FrankBoesing:
\-Os for memory\_copy would work, too.

FrankBoesing:
No -O1 is better. With -Os, it crashes with gcc 5 when using -Os for the \*.ino :)

FrankBoesing:
GCC10 _still_ crashes (sometimes??) with O1

It's needed to disable the interrups as mentioned here:

https://github.com/PaulStoffregen/cores/pull/601#issuecomment-907595637

PR #601, opened 2021-08-27
memory_copy, memory_clear: inline assembler

FrankBoesing:
tested with gcc5, gcc10

### DIFF
--- a/teensy4/startup.c
+++ b/teensy4/startup.c
@@ -52,7 +52,7 @@ void startup_default_early_hook(void) {}
 void startup_early_hook(void)		__attribute__ ((weak, alias("startup_default_early_hook")));
 void startup_default_late_hook(void) {}
 void startup_late_hook(void)		__attribute__ ((weak, alias("startup_default_late_hook")));
-__attribute__((section(".startup"), optimize("no-tree-loop-distribute-patterns")))
+__attribute__((section(".startup")))
 void ResetHandler(void)
 {
 	unsigned int i;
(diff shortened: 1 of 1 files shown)


FrankBoesing:
Hi Paul.
There is something really strange going on. At the  moment, I think there is a combination of problems.
I'm not exactly sure what happens - the generated .lst looks OK to me. It has to do with inlining / cloning etc.
However - the assembler in this PR is OK. Its the way GCC integrates it into the rest of the code.
AS -O1 does nothing dangerous and only very basic optimizations, i'd leave it as it is for now. It's stable in all tests i ran.
Worst case, we could use -O0, someday, to be sure.

FrankBoesing:
Could it be an " inopportune time " interrupt? Not sure.. are the interrupt enabled at the time the copys happen?
In this case I'd add a disable as early as possible to the code, and enable it somewhere after vtor has been written.

FrankBoesing:
Another observation: As first instruction, GCC10 pushes all registers to the stack (before the stackpointer has been written)

So, here is a sum of my suggested changes:

- change the attributes for ResetHandler to: __attribute__((section(".startup"), naked, noreturn))
- as first instruction disable interrupts
- enable them after VTOR has bee written
- keep -O1 for the memcopys

Let me know if you want a PR. I'd do it know, because i'f someday occurs a problem in ResetHandler, nobody will remember...

FrankBoesing:
(I've tested this here, again, with all optimization flags and GCC 5 and GCC 10 and it works in any case)

FrankBoesing:
Last hint: If you want, you can in addition use the asm in this PR, too - but it somehow needs the -O1, too  (despite the fact that there it is a fixed asm which is not modified.. )

PR #610, opened 2021-09-15
Clear IntervalTimer interrupt even if functions are null (teensy4)

### LINKED ISSUE #409: Should TFLG in IntervalTimer (teensy4) be cleared even if the function is nullptr?
https://github.com/PaulStoffregen/cores/blob/06492c1d11b93879b74d8db586d4acce5f2ea5ea/teensy4/IntervalTimer.cpp#L108

It looks like if the function is `nullptr` then the interrupt won't get cleared. Should the interrupt be cleared anyway?

ssilverman:
Fixes #409.


ssilverman:
I just read this thread in the forum: https://forum.pjrc.com/threads/71929-Timer-mystery

I figure now is a good time to ping on this PR.

PaulStoffregen:
As mentioned on #409, I'm not planning to merge this unless a test case or user report shows it can result in wrong behavior in practice (using the IntervalTimer public API) or if the performance impact is carefully analyzed / benchmarked.

PR #611, opened 2021-09-15
Fix HardwareSerial TX and RX FIFO size calculation (teensy4)

### LINKED ISSUE #410: TX and RX FIFO size calculation in HardwareSerial.cpp (teensy4) may need correction
https://github.com/PaulStoffregen/cores/blob/041bfbe9ed85b13f1a833b6c81fe2cd54d1b1ac6/teensy4/HardwareSerial.cpp#L155

Strictly speaking, the TX and RX FIFO sizes should look like this:
```c++
uint16_t tx_fifo_size = (1 << (((port->FIFO >> 4) & 0x7) + 1));
if (tx_fifo_size == 2) tx_fifo_size = 1;
uint16_t rx_fifo_size = (1 << (((port->FIFO >> 0) & 0x7) + 1));
if (rx_fifo_size == 2) rx_fifo_size = 1;
```

ssilverman:
Fixes #410.

### DIFF
--- a/teensy4/HardwareSerial.cpp
+++ b/teensy4/HardwareSerial.cpp
@@ -170,10 +170,26 @@ void HardwareSerial::begin(uint32_t baud, uint16_t format)
 	attachInterruptVector(hardware->irq, hardware->irq_handler);
 	NVIC_SET_PRIORITY(hardware->irq, hardware->irq_priority);	// maybe should put into hardware...
 	NVIC_ENABLE_IRQ(hardware->irq);
-	uint16_t tx_fifo_size = (((port->FIFO >> 4) & 0x7) << 2);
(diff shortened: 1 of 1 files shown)


FrankBoesing:
(48 Bytes) saved.

Also, please think about moving the whole func to FLASHMEM

KurtE:
I would probably just go with what Frank said... As I don't see anyway to set it and I don't think the hardware is changing any time soon.

ssilverman:
I just updated the commit with fixed sizes for the FIFO (4) and the watermark (2). I kept the fixed calculation commented below and added the reference manual source.

ssilverman:
I think this PR will save some bytes.

PR #614, opened 2021-09-20
Reduce memory usage (wasted RAM)

FrankBoesing:


FrankBoesing:
Added two sections for vectors + usb. Results in more savings.
More is not possible.

PR #616, opened 2021-09-21
faster memcpy not needed for GCC 10

FrankBoesing:
Builtin has exactly the same speed.


FrankBoesing:
...probably needed for T3, too. Not part of this PR.

FrankBoesing:
You're right.  Thx.
Paul, please add the paratheses

FrankBoesing:
Too much work. Have deleted the repo already., and it will not be merged any time soon.

Feel free to do so.

Issue #531, opened 2021-01-07
Teensy3 / Teensy4 Stream.timedRead() is private instead of protected


herrfrei:
Issue #388 already described the problem, but I found the declaration if timedRead() as private and not protected as required. The ArduinoHttpClient library as well as khoih.prog/EthernetWebServer that is marked as compatible with Teensy use timedRead() in a  class derived from Client and hence the compiler produces an error.

Solution: declare timedRead() as protected.

ssilverman:
This issue can be closed, as it was fixed in #619.

Issue #388, opened 2019-08-21
Teensy35 Stream.timedRead() is protected


paulhayes:
This method is called by the following library: ArduinoHttpClient @ src\HttpClient.cpp:576:17

The method is public in the main arduino library and other compatables, shouldn't the Stream class be symmetrical across the different platforms? Could you modify this to match arduino framework?

ArduinoHttpClient is listed as being compatible with Teensy, and it is except for this one methods protection level.

FrankBoesing:
It is protected here: https://github.com/arduino/ArduinoCore-avr/blob/master/cores/arduino/Stream.h#L54 too

herrfrei:
I have the same problem, but in the code for Teensy3 and Teensy4 the method in declared private, not protected! Hence the different HttpClient libraries wont't compile.

PR #619, opened 2021-10-02
Stream::timedRead protected instead of private.

KurtE:
Probably not a big issue.  But came up again on forum:
https://forum.pjrc.com/threads/68345-TimedRead-is-private-in-Stream-h-for-Teensy-4

And there have been two issues on it:
https://github.com/PaulStoffregen/cores/issues/531
https://github.com/PaulStoffregen/cores/issues/388

I did a quick check in sources and found it was protected in all three of the Arduino code bases (avr, sam, samd), also with ESP, also with OpenCM/OpenCR...

There are some other changes in the Arduino ones which we could do but are not in the ESP sources

Their method:     int peekNextDigit(LookaheadMode lookahead, bool detectDecimal); // returns the next numeric digit in the stream or -1 if timeout

is different than the ones we have.  The ones we have are now marked private...  and they also added some defines:
enum LookaheadMode{
    SKIP_ALL,       // All invalid characters are ignored.
    SKIP_NONE,      // Nothing is skipped, and the stream is not touched unless the first waiting character is valid.
    SKIP_WHITESPACE // Only tabs, spaces, line feeds & carriage returns are skipped.
};

I have not yet made those changes to peekNextDigit

Probably not a big priority, but since I took a quick look thought I would make the change and PR...

Compiled program on T4.1 and T3.6

### DIFF
--- a/teensy3/Stream.h
+++ b/teensy3/Stream.h
@@ -62,12 +62,13 @@ class Stream : public Print
 	void clearReadError() { setReadError(0); }
   protected:
 	void setReadError(int err = 1) { read_error = err; }
-	unsigned long _timeout;
-  private:
-	char read_error;
 	int timedRead();
 	int timedPeek();
 	int peekNextDigit();
+
+	unsigned long _timeout;
+  private:
+	char read_error;
 };
 
 #endif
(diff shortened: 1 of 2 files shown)


ssilverman:
Fixes #531.

PR #624, opened 2021-11-05
Data timing window holes

KurtE:
@PaulStoffregen  @mjs513:

We were getting partial transfers...   And other issues in MTP whose code was based off of RAWHID code. 

Looked at Sermu and it protected earlier in the rx_queue_transfer...  So it did not grab the index before it disabled the interrupts. So something might change during that window fo time. 
Updated rawhid as well, as well, someone has or will hit it there as well

There may still be some issues in the MTP code, but this appears to me to help some.

### DIFF
--- a/teensy4/usb_mtp.c
+++ b/teensy4/usb_mtp.c
@@ -90,10 +90,10 @@ void usb_mtp_configure(void)
 
 static void rx_queue_transfer(int i)
 {
+	NVIC_DISABLE_IRQ(IRQ_USB1);
 	void *buffer = rx_buffer + i * MTP_RX_SIZE_480;
 	arm_dcache_delete(buffer, rx_packet_size);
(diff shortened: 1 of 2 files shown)


FrankBoesing (commented):
Oh my.. is that needed, and is there no better way? Disabling the interrupts during the loop of cache handling is pretty suboptimal ( And exactly for what reason is it needed? )
With all these diables the interrupt latency gets worse and worse...

FrankBoesing:
Want to see the #define of ESP?

#define sei()  
#define cli()  
#define interrupts() sei()  
#define noInterrupts() cli()

THAT's optimal... :)

FrankBoesing:
And yes I know, your code disables one interrupt only...

KurtE:
does not appear to address problem

PR #626, opened 2021-11-15
T4. usb_mtp.c - on RX look at actual count of bytes transferred and use it

KurtE:
@PaulStoffregen  @mjs513 -

We were running into a case where the host would send us a 0 length packet if on a SendObject the file fit exactly in full packets... Then it would send one that said no more data on that file.

Our code did not look in the transfer->status field to extract the count transferred.
The code now extracts that and only copies that many bytes and returns it to caller.

This helps clear up some of the issues and crashes we have been seeing. 

I am assuming you would prefer the simple fix here and not go off my other branch which added in support
for RawHid512... Although the tests using it were sort of nice as with it you do get 8 times the data throughput

### DIFF
--- a/teensy4/usb_mtp.c
+++ b/teensy4/usb_mtp.c
@@ -36,14 +36,13 @@
 //#include "HardwareSerial.h"
 
 #include "debug/printf.h"
-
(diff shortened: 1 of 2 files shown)


FrankBoesing:
Thank you, Kurt.

PR #629, opened 2021-11-21
Rawhid 512 support for T4.x

KurtE:
@PaulStoffregen ... Also @mjs513 as you were playing along as well...

This is just an experiment, to see how well a version of RAWHID wouild work that supported full 512 byte transfers on T4.x boards.

Works pretty well.  i.e. you get 8 times the throughput of data.

This is discussed in the forum thread: https://forum.pjrc.com/threads/68620-T4-x-support-for-Raw-HID-512-wonder-if-it-makes-sense-to-add-to-system?highlight=rawhid512

You can setup to build either way.  I did it to be able to switch by adding new section to boards.txt or in testing boards.local.txt
```
teensy41.menu.usb.rawhid512=Raw HID 512
teensy41.menu.usb.rawhid512.build.usbtype=USB_RAWHID512
teensy41.menu.usb.rawhid512.fake_serial=teensy_gateway

teensyMM.menu.usb.rawhid512=Raw HID 512
teensyMM.menu.usb.rawhid512.build.usbtype=USB_RAWHID512
teensyMM.menu.usb.rawhid512.fake_serial=teensy_gateway

teensy40.menu.usb.rawhid512=Raw HID 512
teensy40.menu.usb.rawhid512.build.usbtype=USB_RAWHID512
teensy40.menu.usb.rawhid512.fake_serial=teensy_gateway
```

I was testing using at test program that built both for windows and Ubuntu... Have not tried Mac yet...

Code is up at: https://github.com/KurtE/RawHid

Based off of your zip file on main website.

I also have test code that I may do PR on in USBHost_t36 that supports this as well, so you have T4 to T4 with large buffers.

There are probably still some issues to work out  Like on Ubuntu I detect which one the Teensy is configured for by looking at the Endpoint TX and RX sizes.  On windows I am currently looking at the HID definition. This may be wrong, as well if you connect to a USB that does not support high speed it will connect at 12mbs and at that point the connection will be back to 64 byte packets.

So again mostly playing, but thought I would do a PR to you, to see if this is something you are interested in.

Extracted from my WIP branch rawhid_512.  
That other branch had debug code I added to core to print crashreport data to Serial4 if configured for it at the time of the crash (debug\printf.h)... Did that as the crash for overwriting memory was such that the crashreport data was not valid after reboot, so did not display  Could PR that as well if interested (startup.c)


Oskar888:
@PaulStoffregen Why didn't you implement this? this would be a great solution, especially since teensy does not support webusb, so only webhid remains.

PR #637, opened 2022-03-01
Update Time.cpp

skunktrading:
Without the proposed change, the calculated time skips 01 March as the leap year determination is incorrect.


skunktrading:
This PR is redundant since https://github.com/PaulStoffregen/cores/pull/742 was merged

PR #648, opened 2022-06-07
T3.x - Add Serial+MTP

KurtE:
@mjs513 @PaulStoffregen 

I used the stuff out of MTP_T4 stuff, but modified it to use the new PID.  Also changed the product name to match what you have for the T4.x which did not include the word serial. 

Added the T3.x Serial+MTP to the usb_desc.h

did a rudimentary test with the simple MTP + MSC test that ran on T4.x including MMOD, and tested it on T3.6 with usb host cable with a SSD plugged into it.


PaulStoffregen:
Looks like PRODUCT_ID is defined twice.  The RawHID defines should not be needed.

KurtE:
Ooops, meant to remove that old PID line, ... So pushed up without it up.  Also removed the old HID stuff from the define

PR #651, opened 2022-08-24
Add AudioMemoryNoForceUpdate()

h4yn0nnym0u5e:
Although most users will need audio updates working, for development purposes it's sometime helpful to disable them and run updates from within the sketch. This feature adds that capability back in, which was removed in commit 503cd9f

### DIFF
--- a/teensy3/AudioStream.cpp
+++ b/teensy3/AudioStream.cpp
@@ -61,7 +61,7 @@ AudioConnection* AudioStream::unused = NULL; // linked list of unused but not de
 
 // Set up the pool of audio data blocks
 // placing them all onto the free list
-void AudioStream::initialize_memory(audio_block_t *data, unsigned int num)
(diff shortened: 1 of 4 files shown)


PaulStoffregen:
Why expand the public API to support this test scenario?

A test program can create a class which calls update_setup() from its constructor.  Then it is responsible for updates.  But if you want to manually update the library, just create a public function which calls update_all(), rather than calling it from hardware-based timing.

h4yn0nnym0u5e:
Why? To restore functionality to existing code, which _will_ be broken by this change, is a fairly easy to use way.

But as I've noted on the forum, I think there will potentially be other nasty surprises arising from the change you put in, which this PR doesn't address, so there's not a lot of point pursuing it as-is.

PaulStoffregen:
Let's find those nasty surprises...

PR #655, opened 2022-09-14
Fix for Conflict with abs/round with stdlib (toolchain 11.3.1 issue)

mjs513:
@PaulStoffregen  -- @flybrianfly

When testing some other libraries that we used [Eigen](https://github.com/bolderflight/eigen) and a [Ublox ](https://github.com/bolderflight/ublox) I was receiving a ton of errors all related to using abs and round defined in wiring.h.  Noticed that there was an open issue with brtaylor's Eigen library for the same thing when using toolchain 7,x.x.  So made the changes in wiring and it did indeed clear up the issue.  

See issue: https://forum.pjrc.com/threads/71074-Teensyduino-1-58-Beta-2?p=312709&viewfull=1#post312709

The PR makes the change for Teensy4 and Teensy3.

### DIFF
--- a/teensy3/wiring.h
+++ b/teensy3/wiring.h
@@ -129,29 +129,38 @@ constexpr auto max(A&& a, B&& b) -> decltype(a < b ? std::forward<A>(a) : std::f
 #define SERIAL  0
 #define DISPLAY 1
 
+//commented out as a result of moving to 11.3.1 toolchain
 // undefine stdlib's abs if encountered
-#ifdef abs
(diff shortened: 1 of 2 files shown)


PaulStoffregen:
Ok, let's give this a try in 1.59-beta3 and see if any ancient Arduino code breaks...

mjs513:
Thanks. Hopefully not.  Seems to be cropping up more now for some reason and seems to be only with round and abs so far.

PR #656, opened 2022-09-18
IntervalTimer Renovation, std::function replacement

luni64:
**Please note: This requires the new toolchain**

## IntervalTimer with lightweight std::function like interface

I know (and fully understand) that you currenty don't want to pull that in. But, since I'm currently playing with this stuff I thought it might be a good idea to place the PR anyway, so that you can have a look at it whenever there is time. In case this version is "too much modern c++" for your taste,  here a link to another implementation where I only changed the really necessary stuff: https://github.com/luni64/staticFunctional/tree/main/examples/04_teensy_intervaltimer

**Usage examples** 
(of course standard stuff like calling free void(*)() functions works as expected):

Demonstrating a **capturing lambda** expression. All 4 timers invoke the same callback and pass the timer number to it. 
```c++
IntervalTimer t[4];

void onTimer(int ch)
{
    Serial.printf("Timer #%d, called at %dms\n", ch, millis());
}

void setup()
{
    for (int i = 0; i < 4; i++)
    {
        t[i].begin([i] { onTimer(i); }, random(200'000, 1'000'000));
    }
}
void loop() {
}
```

**Embedding** a Intervaltimer in a user class:
```c++
class blinker
{
 public:
    void begin(unsigned _pin, float frequency) {
        pin = _pin;
        pinMode(pin, OUTPUT);
        t.begin([this] { doBlink(); }, 0.5E6 / frequency); // initialize the embedded timer using a lambda capturing 'this'
    }
    void doBlink() { 
        digitalToggleFast(pin);
    }
 protected:
    unsigned pin;
    IntervalTimer t;
};

blinker b1,b2;

void setup()
{
    while (!Serial) {}

    b1.begin(LED_BUILTIN, 10); // 10 Hz blinking LED
    b2.begin(0, 50'000);       // 50kHz signal on pin 0
}
void loop() {
}
```
Using **time literals**
```c++
using namespace std::literals; // we want to use time literals (ms, ns ....)

IntervalTimer t1, t2, t3;

void onTimer(const char* msg){
    Serial.printf("%s timer called at %d\n", msg, millis());
}

void setup(){
    t1.begin([] { onTimer("320ms"); }, 320ms);        // conversion of duration types to the
    t2.begin([] { onTimer("2.5s "); }, 2.5s);         // required microseconds is done at compile time,
    t3.begin([] { onTimer("850ms"); }, 800ms + 50us); // no runtime penalty
}
void loop(){
}

```


PaulStoffregen:
Thanks.  Yes, I will need some time to digest this and other stuff on the forum.  I want to spend some time really looking at the generated asm listing for various scenarios.

I do intend to transition to this sort of callback, maybe for 1.58, maybe delayed until 1.59... and build more non-blocking APIs using callbacks.

PaulStoffregen:
Looks like quite a lot of template magic happening in staticFunctional.h.

Is there a source for this with MIT or MIT-like open source license?  It really can't be merged to the core library without attribution and clear indication the author(s) approve of MIT license terms.

luni64:
> Looks like quite a lot of template magic happening in staticFunctional.h.

Indeed, my [CallbackHelper ](https://github.com/luni64/CallbackHelper)is much simpler (but also clumsier). A part of the complexity  comes from the adaption to c++11 build systems. The original code (c++20?)  was a bit shorter. 

> Is there a source for this with MIT or MIT-like open source license? 

Code is from a forum post. So it is not licenced. Here I asked the author (Columbo) if I can use his code for a MIT licenced library. He agreed: https://www.c-plusplus.net/forum/topic/344451/std-function-und-std-bind-durch-einfachere-eigene-variante-ersetzen-signal-slot/25.  
Since his code, by his allowance, is now in my (MIT licenced) library, I'd say fomally you could simply refer to that library. But it might be more friendly to ask him directly :-). 

BTW: I had a quick look at the assembly of `pit_isr` and compared the orignal code with the new code. I'd say, the main difference is an additional indirection to invoke the callback. This comes from the fact that the call is done through a virtual base class.  So, it needs to be called through the vtable. Nevertheless, suprisingly little overhead.

lycheeblossom:
I’ll try to keep an eye out for this when you’re ready to begin focusing on it. I’ve had a mind to write non-blocking (along with more direct IMXRT) versions of a variety of libraries for quite a while… and a particularly harrowing period of crunch at work is finally over.

PaulStoffregen:
This article discusses some of the trade-offs involved.

https://quuxplusone.github.io/blog/2019/03/27/design-space-for-std-function/

My first impression (could be wrong, haven't worked with this code yet) is this looks like "owning".  For use with non-blocking APIs on Teensy, I definitely want a non-owning approach.

luni64:
Thanks for the article, that and the linked article about type erasure are quite interesting and well written information,  I'll dive deeper into it later.  

> For use with non-blocking APIs on Teensy, I definitely want a non-owning approach.

I currently don't understand the reasoning behind that. Can you help me to understand how "non owning" is related to "non blocking APIs" and why it would be "bad"?

luni64:
The blog article links to a[ inplace_function](https://github.com/WG21-SG14/SG14/tree/master/SG14) which is proposed for c++21 by some [SG14 ISO group](https://lists.isocpp.org/mailman/listinfo.cgi/sg14).  

Inplace_function has  same functionality as the staticFunctional code I uploaded yesterday. But, having this code suggested at this official(?) ISO SG14 repo is kind of reassuring :-) 

I tried the linked code and it works out of the box (well, I had to disable throwing an exception in case one tries to invoke an empty function, might be a good idea to trigger a crash instead). 

I uploaded it to the PR if you want to try it. You can switch between `std::function`, `stdex::inplace_function` and `staticFunctional::function` in `intervaltimer.h` by setting FUNCTION_TYPE to 0, 1 oder 2

Of course this is also an "owning" implementation. But honestly I still don't get what would be wrong about this.

lycheeblossom:
Thank you! I’ll review the proposed implementation today and toy around with it when I have a moment during an evening (or two) this week.

luni64:
I don't know if I undestood your non blocking libraries idea correctly. Anyway, I tried if I would run into a wall using the owning inplace_function for some super simple non blocking proof of concept code.

I implemented an asyncFunction handler which is basically just an array of <bool()> functions. 
In its tick function (typically called from yield) the handler works through its list of stored functions. The functions will be called on each tick until they return false. In this case the hander removes them from the list. 

A user class can register a worker function which will then be invoked from yield, i.e. asynchronous in the background. Here the code: 
```c++
#include "Arduino.h"

using cb_t = stdext::inplace_function<bool(void)>;

class asyncHandler_t
{
 public:
    // looks for a free slot to copy the callback to.
    bool add(cb_t callback) // registers a callback to the handler
    {
        for (auto& func : asyncFunctions)   { // find a free slot 
             if (func == nullptr)  {
                  func = callback;
                  return true;
             }
         }
        return false;
    }

    // call this as often as possible, e.g. from yield.
    void tick() {
        for (int i = 0; i < 20; i++) { // invoke all stored callbacks. Could of course only invoke one callback per tick...        
            auto& f = asyncFunctions[i];
            if (f) {
                bool done = f(); // invoke callback
                if (done) {
                    f.~inplace_function();       // call destructor in case f wraps a callback which needs to be destructed 
                    asyncFunctions[i] = nullptr; // remove the callback from the list
                }
            }
        }
    }
 protected:
    cb_t asyncFunctions[20];  // 20 slots, 
};

asyncHandler_t asyncHandler;

//=============================================================================================
// User class

class Test
{
 public:
    Test(const char* _whoAmI) { // identifier
        whoAmI = _whoAmI;
    }

    void doSomethingLong(){ // performs some long task in pieces, called from yield    
        asyncHandler.add([this] { return worker(); });   // register our non static worker to the handler
    }
 protected:
    bool worker(){ // This worker splits its task i 3 subtasks. As long as it returns false it will be called from yield.    
        Serial.printf("%s -->  ", whoAmI);
        switch (state++) {
            case 0:
                Serial.println("Did first task");
                // do some background task
                return false;
            case 1:
                // do something in the background
                Serial.println("Did second task");
                return false;
            case 2:
                Serial.println("Did last task");
                state = 0;
                return true;
            default:
                return true;
        }
    }

    int state = 0;
    const char* whoAmI;
};

bool start = false;

Test test1("First  object");
Test test2("Second object");

void setup(){
    while (!Serial) {}
    if (CrashReport) Serial.println(CrashReport);

    pinMode(LED_BUILTIN, OUTPUT);
    start = true; 

    test1.doSomethingLong();
    test2.doSomethingLong();
    delay(10);
    Serial.println(".....");
    test1.doSomethingLong();
}

void loop()
{
    digitalToggleFast(LED_BUILTIN);
    delay(150);
}
//---------------------------------------------------
void yield()
{
    if (start) // yield could be called before setup...
        asyncHandler.tick(); // work the task list from yield
}
```

It prints:
```
First  object -->  Did first task
Second object -->  Did first task
First  object -->  Did second task
Second object -->  Did second task
First  object -->  Did last task
Second object -->  Did last task
.....
First  object -->  Did first task
First  object -->  Did second task
First  object -->  Did last task
```

PaulStoffregen:
At this moment I'm stalled on dealing with more supply chain problems.  Probably won't be able to do much with software for at least the next few weeks.

PaulStoffregen:
Now that we have inplace_function and IntervalTimer using it, can this PR be closed?

Or in other words, is anything we haven't already done still here?

luni64:
Sure, that was just a demonstration anyway

PaulStoffregen:
Thanks.  I'm just going through lists of issues and forum threads before packaging up 1.59-beta2 with the new callback stuff.

PaulStoffregen:
@lycheeblossom and anyone else following this (and wanting to use C++ lambda callbacks from libraries), the conversation was continued on this forum thread:

https://forum.pjrc.com/threads/70986-Lightweight-C-callbacks?p=323559&viewfull=1#post323559

As of today, version 1.59-beta2 is available with inplace_function class and used inside IntervalTimer.

https://forum.pjrc.com/threads/72572-Teensyduino-1-59-Beta-2

PR #663, opened 2022-11-03
MTP Teensy4 - MTP.loop() would wait about 1ms with no messages

KurtE:
We passed in -0 for TO, but the code was waiting for systeick_miillis_count to change before we would return.  This was very noticeable with test sketches.

Also impacting some other sketches as well.  May also need to update for T3.x but I think that is contained within MTP_Teensy

### DIFF
--- a/teensy4/usb_mtp.c
+++ b/teensy4/usb_mtp.c
@@ -139,7 +139,7 @@ int usb_mtp_recv(void *buffer, uint32_t timeout)
 	while (1) {
 		if (!usb_configuration) return -1; // usb not enumerated by host
 		if (tail != rx_head) break;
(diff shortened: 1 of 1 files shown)


KurtE:
Edit should mention some of the timing information including some from @mjs513 
is up on the forum thread:
https://forum.pjrc.com/threads/71424-SD-mediaPresent()-MTP-Audio-issues?p=315284#post315284

h4yn0nnym0u5e:
I've tested this with the "audio guestbook" as improved by DD4WH (https://forum.pjrc.com/threads/70553-Teensy-4-0-based-Audio-Guestbook), and can report that it's effective in fixing the intermittent lost audio. Recommended for beta test ASAP!

PR #673, opened 2023-01-07
Fix a startup issue with non-default optimisation option

h4yn0nnym0u5e:
For some reason compiling `startup.c` with e.g. -Og causes an issue, resulting in the Teensy never starting up and enumerating, and the user having to press the Program button to get back to bootloader mode. Changing to -O2 optimisation for the `reset_PFD()` function seems to fix the issue.

As discussed in https://forum.pjrc.com/threads/71846-Teensy4-1-optimisation-0g-00-fails-to-start, for example. Fairly sure there are multiple similar reports / threads. A good way to provoke it is to try to use the Audio and TeensyDebug libraries together.

### DIFF
--- a/teensy4/startup.c
+++ b/teensy4/startup.c
@@ -509,6 +509,11 @@ FLASHMEM void usb_pll_start()
 	}
 }
 
+// For some reason compiling this file with e.g. -Og causes an issue, resulting
+// in the Teensy never starting up and enumerating, and the user having to press
+// the Program button to get back to bootloader mode. Changing to -O2 optimisation
+// for this one function seems to fix the issue.
+__attribute__((optimize("O2")))
 FLASHMEM void reset_PFD()
 {	
 	//Reset PLL2 PFDs, set default frequencies:


PaulStoffregen:
Committed a fix.
https://github.com/PaulStoffregen/cores/commit/3f95c8e6f632d76ca20210cd5edef92fa791191d

h4yn0nnym0u5e:
Example of sketch that will cause startup to fail. Arduino IDE settings:
- Teensy 4.1
- Dual Serial
- 600MHz
- Faster
- **GDB: "Use Dual Serial"**

Reproduced using `cores` and `Audio` from github repos as of 2025-04-27.

You must have [TeensyDebug](https://github.com/ftrias/TeensyDebug) in your libraries and installed according to its instructions - it needs to hook into the upload process to launch GDB after Teensy has rebooted (tested with revision 0.0.4). Edit `src/TeensyDebug.h` line 24 to `#define IRQ_DEBUG IRQ_SOFTWARE+1` to avoid clash with Audio interrupt.

Symptoms are that USB will fail to enumerate after the binary has been uploaded; pressing the Program button does work, however.

I tried reading "Configuration of Phase Fractional Dividers EB790" but was none the wiser as to any timing requirements for these registers, but the optimisation does indeed make a subtle change. The requirement _sometimes_ to comment out _one_ of the calls to `reset_PFD()` remains a mystery to me...

[BugStalledStartup.zip](https://github.com/user-attachments/files/19931540/BugStalledStartup.zip)

h4yn0nnym0u5e:
Here's a version _not_ requiring TeensyDebug. Arduino IDE settings:
- Teensy 4.1
- Dual Serial
- 600MHz
- **Optimize: "Debug"**

Otherwise all as above.

[BugStalledStartup.zip](https://github.com/user-attachments/files/19937381/BugStalledStartup.zip)

PaulStoffregen:
Any chance I could talk you into retesting the TeensyDebug test case(s) with the recent 1.60-beta5 version?

https://forum.pjrc.com/index.php?threads/teensyduino-1-60-beta-5.77176/

h4yn0nnym0u5e:
> Any chance I could talk you into retesting the TeensyDebug test case(s) with the recent 1.60-beta5 version?
> 
> https://forum.pjrc.com/index.php?threads/teensyduino-1-60-beta-5.77176/

Sure, I'll take a look - may be a few days as I've got other stuff on this weekend. I _can_ say I've recently been using TeensyDebug a bit, on a rather grubby fork of `cores` which includes other (unrelated AFAIK) mods, along with these changes to `startup.c`. It seems to be fine, even though I'm using the Audio library which historically caused issues. But of course, the problems always were a bit intermittent...

Anyway, I'll give a clean TD1.60b5 a go, and report back.

PR #678, opened 2023-01-14
Change _write() to return -1 if the file is stdin

ssilverman:
Note that ideally, `errno` should be set if the file is `stdin`. I'll create a separate PR for that. Returning zero seems like a reasonable choice for now.

### DIFF
--- a/teensy3/Print.cpp
+++ b/teensy3/Print.cpp
@@ -86,7 +86,11 @@ extern "C" {
 __attribute__((weak))
 int _write(int file, char *ptr, int len)
 {
-	if (file >= 0 && file <= 2) file = (int)&Serial;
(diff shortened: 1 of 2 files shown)


PaulStoffregen:
Keeping the Print class code small and simple is of higher importance than rigorous unix / posix compatibility.

PaulStoffregen:
In fact, I considered testing only if the file int is equal to 1....

ssilverman:
What about just testing for the range 1-2 instead of 0-2, and without my return-negative-one-if-check? Would that be better here? No worries if you think 0-2 is the best choice for you; I can close this if that’s the case.

PaulStoffregen:
Let's just leave it as-is and only revisit this if anyone encounters a significant problem.

PR #680, opened 2023-01-19
Fix UK keyboard compile warning

ssilverman:
Because this keyboard defines KEYCODE_EXTRA00, usb_keyboard.c's unicode_to_keycode() can return a 16-bit quantity.

See:
https://forum.pjrc.com/threads/71926-Button-Press-detector?p=319143&viewfull=1#post319143

### DIFF
--- a/teensy/keylayouts.h
+++ b/teensy/keylayouts.h
@@ -1446,8 +1446,8 @@ extern "C"{
 
 #define SHIFT_MASK		0x0040
 #define ALTGR_MASK		0x0080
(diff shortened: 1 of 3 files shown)


ssilverman:
Note: I didn't modify the Teensy 2 code. I'm hesitant because I'm not certain how much change should be introduced there.

ssilverman:
I changed my mind and changed the Teensy 2 code.

ssilverman:
I just force pushed changing the `KEYCODE_MASK` to 0x7ff. Still figuring out if that's correct.

PR #682, opened 2023-01-25
Ensure all _serialEvent_default definitions and declarations match

### LINKED ISSUE #681: EventResponder: _serialEvent_default not working as expected
I think commit 8b225ef144ef53d57045bfac74401f802337a72e breaks the intended behavior of the `_serialEvent_default` stuff. 

I discovered the following issue when compiling for Teensy 4.1 with gcc 12.2.0 (*): the constant `_serialEvent_default` is set to `0` instead of `1`. (it's the same for the other `_serialEvent`_X_`default` constants.)

The global variable `_serialEvent_default` is weak/public defined in EventResponder.cpp as `extern const uint8_t _serialEvent_default __attribute__((weak)) PROGMEM = 0;` and strong in serialEvent.cpp as `const uint8_t _serialEvent_default PROGMEM = 1;`.
From my understanding `_serialEvent_default` should result in a constant set to `1` to make the whole thing work correctly. Otherwise for example
https://github.com/PaulStoffregen/cores/blob/6d6c215501509ee49cdd222992d1f788ab66a341/teensy4/yield.cpp#L61 is never executed.

I guess the extern (public) definition was introduced as a workaround for [gcc bug 83271](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83271)? 
But to make the workaround complete (and not changing the behavior of the `yield()` implementation), you have to define `const uint8_t _serialEvent_default PROGMEM = 1;` as well as extern (public) to make the override of the weak definition work. Therefore `extern` must be added to all the `_serialEvent`_X_`default` definitions in `serialEvent`_(USB)X_`.cpp` for teensy3 and teensy4. 

--
(*) It's the same when compiling with gcc 5.4.1 and it should be independent of the gcc version.

ssilverman:
Make sure they all have 'extern const' so that the weak versions get properly overridden.

See: https://github.com/PaulStoffregen/cores/issues/681

Fixes #681

### DIFF
--- a/teensy3/serialEvent.cpp
+++ b/teensy3/serialEvent.cpp
@@ -3,4 +3,4 @@
 void serialEvent() __attribute__((weak));
 void serialEvent() {
 }
-const uint8_t _serialEvent_default PROGMEM = 1;
+extern const uint8_t _serialEvent_default PROGMEM = 1;
--- a/teensy4/serialEvent.cpp
+++ b/teensy4/serialEvent.cpp
@@ -3,4 +3,4 @@
 void serialEvent() __attribute__((weak));
 void serialEvent() {
 }
-const uint8_t _serialEvent_default PROGMEM = 1;
+extern const uint8_t _serialEvent_default PROGMEM = 1;


ssilverman:
I tested `_serialEvent_default` without the change (i.e. missing `extern`) and confirmed that its value is 0, and then with the change and confirmed that its value is 1. I simply printed the value with `Serial.println(_serialEvent_default)` at the top of `yield()` in _yield.cpp_.

tsandmann:
I think you missed to change `_serialEvent1_default` to `_serialEvent8_default` from HardwareSerial**1**.cpp to HardwareSerial**8**.cpp and serialEvent**1**.cpp to serialEvent**8**.cpp for teensy4. (For teensy3 it's only up to 6).

ssilverman:
I'm not sure those are required. I tried building programs that use `Serial1` through `Serial8` with LTO, but got no errors. I only made, in commit 8b225ef144ef53d57045bfac74401f802337a72e, the necessary changes to get LTO to link without complaints.

ssilverman:
I just forced pushed some changes that remove the USB1 and USB2 changes to keep this localized.

tsandmann:
> I'm not sure those are required. I tried building programs that use `Serial1` through `Serial8` with LTO, but got no errors. Unless... I need to use them in a specific way? It might also have something to do with the fact that Serials 1-8 aren't used in _yield_? I only made, in commit [8b225ef](https://github.com/PaulStoffregen/cores/commit/8b225ef144ef53d57045bfac74401f802337a72e), the necessary changes to get LTO to link without complaints. Thoughts?

That's true, as long as these variables are not used, it's not an issue. As far as I understand it, they are used in https://github.com/PaulStoffregen/cores/blob/6d6c215501509ee49cdd222992d1f788ab66a341/teensy4/HardwareSerial.cpp#L234, but I don't know, why that's not an issue for the compiler in that case.

I don't get the benefit of these `_serialEvent_default` variables at all, so I don't know if you need to use them in a specific way. But I guess @PaulStoffregen does 🙂.

ssilverman:
Dunno. But in summary, here’s my proposal, to make it easier on Paul; either:
1. Revert my LTO fix (8b225ef144ef53d57045bfac74401f802337a72e), or
2. Take the fix in this PR so that yield works the same.

ssilverman:
Just pinging on this because I think it's an important change/fi; it puts back how `yoeld` worked before that LTO "fix". See https://github.com/PaulStoffregen/cores/pull/682#issuecomment-1407494793 for two possible options.

ssilverman:
I just realized that I’m probably explaining this PR wrong and wanted to add this for posterity:
This PR doesn’t fix an LTO bug. It fixes a non-LTO problem that the original LTO fix attempt introduced.

PaulStoffregen:
I removed _serialEvent_default as part of refactoring the yield() optimization.

Closing this pull request, as it no longer applies.

PR #687, opened 2023-03-13
correct TIMSEL field width in FlexIO SHIFTCTL register

squidcc:
The definition of FLEXIO_SHIFTCTL_TIMSEL has the wrong width (only 2 bits when it should be 3) which causes issues when trying to make a shift buffer use timer4/5/6/7. Verifiable on page 2992 of IMXRT1060RM_rev3.pdf.


PaulStoffregen:
Thanks!

PR #689, opened 2023-03-16
Fix a bunch of "unused parameter" warnings

ssilverman:
Discovered when building with -Wall and -Wextra (those include
-Wunused-parameter).


ssilverman:
I'm a fan of when compilation is less noisy when all the warnings (or at least most of the warnings) are turned on.

PaulStoffregen:
It is a bit verbose, but seems ok

Issue #690, opened 2023-03-16
Should this line in usb.c fall through to the next case? ("Experimental" case)


ssilverman:
https://github.com/PaulStoffregen/cores/blob/1aac6ba8783504df8561f9bd35e6032ece719086/teensy4/usb.c#L536

ssilverman:
Fixed by #691.

PR #691, opened 2023-03-16
Fix two "fallthrough" warnings

ssilverman:
Disovered with -Wextra (-Wimplicit-fallthrough).


ssilverman:
Depending on the resolution to #690, this may or may not need to remove one of the two additions.

ssilverman:
Fixes #690.

Issue #660, opened 2022-10-03
New (valid) warnings raised by GCC 11 in v1.58 beta 2


tttapa:
Hi Paul,

When compiling using GCC 11 from Teensyduino 1.58 beta 2 with `-Wextra` enabled, I get the following two warnings in the Teensy core files. (I usually compile with `-Wall -Wextra -Werror` in the CI builds for my libraries.)  
I believe that these warnings are indeed valid.
```cpp
In file included from /home/pieter/.arduino15/packages/teensy/hardware/avr/1.58.0-beta2/libraries/SPI/SPI.h:23:
/home/pieter/.arduino15/packages/teensy/hardware/avr/1.58.0-beta2/cores/teensy3/DMAChannel.h: In copy constructor 'DMASetting::DMASetting(const DMASetting&)':
/home/pieter/.arduino15/packages/teensy/hardware/avr/1.58.0-beta2/cores/teensy3/DMAChannel.h:405:25: error: implicitly-declared 'constexpr DMASetting& DMASetting::operator=(const DMASetting&)' is deprecated [-Werror=deprecated-copy]
  405 |                 *this = c;
      |                         ^
/home/pieter/.arduino15/packages/teensy/hardware/avr/1.58.0-beta2/cores/teensy3/DMAChannel.h:403:9: note: because 'DMASetting' has user-provided 'DMASetting::DMASetting(const DMASetting&)'
  403 |         DMASetting(const DMASetting &c) {
      |         ^~~~~~~~~~
```
Here, the implicitly declared copy assignment operator `operator=(const DMASetting &)` is used instead of the user-defined assignment operator `operator=(const DMABaseClass &)`. Although I'm not familiar with the specifics, I believe this might be a mistake, because `copy_tcd` is not called in this case.  
If it is indeed a mistake, I'd recommend explicitly implementing or `=delete`ing `operator=(const DMASetting &)`. If it's not a mistake, I think you should be able to explicitly use `DMASetting &operator=(const DMASetting &) = default`. 

With USB type set to MIDI, you also get this warning about an invalid function pointer cast:
```cpp
In file included from /home/pieter/.arduino15/packages/teensy/hardware/avr/1.58.0-beta2/cores/teensy3/WProgram.h:62,
                 from /tmp/arduino_build_737688/pch/Arduino.h:6:
[...]

PR #693, opened 2023-03-17
Fix an "implicitly-declared copy assignment operator" warning

ssilverman:
Discovered with -Wextra (-Wdeprecated-copy).

Addresses part of #660.

I _think_ I've done this correctly, but please triple-check my changes. (@luni64 I'd love your opinion here too.) I believe the behaviour changes because, when it uses the implicitly-declared copy assignment operator, `copy_tcd()` isn't called, but with my changes, it _is_ called.


ssilverman:
Making this a draft for now.

ssilverman:
This commit also makes sure each copy constructor has a sibling `operator=()`.

luni64:
Here is my assessment as requested by @ssilverman: 
 
By default, if not provided or deleted by the user, the compiler automatically generates both a `copy constructor` and the corresponding `assignment operator`. If the class has a user-defined `copy constructor` but no user-defined `assignment operator`, auto-generating the missing `assignment operator` is considered `dangerous` and is deprecated. 

This results in the following very explicit warning message:
```c++
.../DMAChannel.h: In copy constructor 'DMASetting::DMASetting(const DMASetting&)':
.../DMAChannel.h:524:25: warning: implicitly-declared 'DMASetting& DMASetting::operator=(const DMASetting&)' is deprecated [-Wdeprecated-copy]
  524 |                 *this = c;
      |                         ^
.../DMAChannel.h:521:9: note: because 'DMASetting' has user-provided 'DMASetting::DMASetting(const DMASetting&)'
  521 |         DMASetting(const DMASetting &c)
      |         ^~~~~~~~~~
```
Since deleting the `assignment operator` is not an option here, one should define a corresponding user `assignment operator` as @ssilverman  proposes.

PR #694, opened 2023-03-17
Fix some "cast between incompatible function types" warnings in usb_midi

ssilverman:
Discovered with -Wextra (-Wcast-function-type).


ssilverman:
This also addresses some of #660.

PR #696, opened 2023-03-20
Change internal Print::write_error to int from char

### LINKED ISSUE #697: Print::setWriteError(int) loses information
`Print::setWriteError(int)` loses information because it translates the `int` argument to an internal `char`. This means that `getWriteError()` won't always return the expected value.

ssilverman:
This prevents information from being lost.

Fixes #697.

### DIFF
--- a/teensy3/Print.h
+++ b/teensy3/Print.h
@@ -118,7 +118,7 @@ class Print
   protected:
 	void setWriteError(int err = 1) { write_error = err; }
   private:
-	char write_error;
+	int write_error;
 	size_t printFloat(double n, uint8_t digits);
 #ifdef __MKL26Z64__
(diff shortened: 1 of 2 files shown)


ssilverman:
All these use `int` internally:
1. https://github.com/arduino/ArduinoCore-avr/blob/eabd762a1edcf076877b7bec28b7f99099141473/cores/arduino/Print.h#L40
2. https://github.com/esp8266/Arduino/blob/734defb87dc09104856c21fa4f2a6bb912f7330a/cores/esp8266/Print.h#L38
3. https://github.com/espressif/arduino-esp32/blob/0d84018d969309addacbcc3e3782c1fadc95fbc8/cores/esp32/Print.h#L37

PR #699, opened 2023-04-09
Replace begin / update functions with different signature by a simple c++17 template

luni64:
I played around with some new c++17 features which enable e.g. replacing all your begin/update functions with similar signatures (int, uint, long, float double etc) by a simple template using the new `constexpr if`. It is guarateed to be evaluated at compile time and makes template syntax much more friendly to the eyes :-)

PS: needed to remove the c-linkage, was this intended anyway? 

PPS: There are some whitespace changes due to my autoformatter. Let me know if you consider pulling this, then I'd change it to stick to your usual formatting style.

### DIFF
--- a/teensy4/IntervalTimer.h
+++ b/teensy4/IntervalTimer.h
@@ -33,155 +33,97 @@
 
 #include <stddef.h>
 #include "imxrt.h"
+#include <type_traits>
+
 #if TEENSYDUINO >= 159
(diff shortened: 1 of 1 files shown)


PaulStoffregen:
Can you please redo this without changing all the white space?

luni64:
Replaced by #700

PR #707, opened 2023-07-01
Add some checksum

Defragster:
Stored RAM2 values give false breadcrumb printing unrelated to a CrashReport

Many times a CrashReport on first run will cause BreadCrumb printing of unrelated garbage

This quick fix writes bitmask to checksum and verifies they are equal before printing.

WARNING for TEST there is a BUGBUG DEBUG else{} in the code that shows as below from extra code in CrashReport.cpp
If you see this as a worthy addition I can resubmit with lines 218-223 removed
Without that exclusion code it would read the bitmask bit and display values.
Print below shows the values stored in the T_4.0 here on upload and sample crash caused.

CrashReport:
  A problem occurred at (system time) 0:40:53
  Code was executing from address 0xFFFFFFFE
  CFSR: 83
	(IACCVIOL) Instruction Access Violation
	(MMARVALID) Accessed Address: 0x0 (nullptr)
	  Check code at 0xFFFFFFFE - very likely a bug!
	  Run "addr2line -e mysketch.ino.elf 0xFFFFFFFE" for filename & line number.
  HTSR: 40000000
	(FORCED) Forced Hard Fault
  Temperature inside the chip was 36.89 °C
  Startup CPU clock speed is 600MHz
  Reboot was caused by auto reboot after fault or bad interrupt detected
**FALSE Panic Breadcrumb printing! BUGBUG DEBUG >>bitmask==45C48D85
FALSE Panic Breadcrumb printing! BUGBUG DEBUG >>checksum==145D6C65**

### DIFF
--- a/teensy4/CrashReport.cpp
+++ b/teensy4/CrashReport.cpp
@@ -199,7 +199,7 @@ size_t CrashReportClass::printTo(Print& p) const
 	  asm volatile ("dsb":::"memory");
 	  while (1) asm ("wfi");
   }
-  if (bc->bitmask) {
+  if (bc->bitmask && bc->bitmask == bc->checksum) {
     for (int i=0; i < 6; i++) {
       if (bc->bitmask & (1 << i)) {
         p.print("  Breadcrumb #");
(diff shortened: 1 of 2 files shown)


Defragster:
Opps - these DEBVUG lines at 218 should not have been included:
`    else if (bc->bitmask) {
      p.print("FALSE Panic Breadcrumb printing! BUGBUG DEBUG >>bitmask==");
      p.println(bc->bitmask, HEX);
      p.print("FALSE Panic Breadcrumb printing! BUGBUG DEBUG >>checksum==");
      p.println(bc->checksum, HEX);
    }`

Defragster:
Not tested yet, but looks good Paul. Sorry for including the BUGBUG in the PR - but nobody else had seen it an wasn't sure if testing would be needed to see the issue. Adding true checksum is better - except for the added cycles on each breadcrumb update.

Defragster:
Single test on cold start no longer give bogus breadcrumbs with edited cores. Using BUGBUG fix on 1.58 IDE 2 gives those false prints

PR #114, opened 2016-02-10
Add OUTPUT_OPENDRAIN, INPUT_PULLDOWN for pinMode()

Xenoamor:
I'm unfamiliar with the MKL26Z64VFT4 and am unsure whether or not this functionality is supported by it.

I also can't seem to edit keywords.txt to make the new #defines appear blue in the Arduino IDE like OUTPUT and INPUT. Only orange

PR #708, opened 2023-07-09
enable prefetching when accessing PSRAM over AHB

A-Dunstan:
The prefetch feature of FlexSPI2 is not being activated in FLEXSPI2_AHBCR, reducing performance when PSRAM is accessed sequentially.
This commit turns it on and allocates all prefetch buffer space to RXBUF3 which will be used for all AHB masters (CPU/DMA/DCP/others).

As an example of possible performance improvement, this reduces the time for the psram test sketch from 36.54 seconds to 29.99 seconds (1x 8MB PSRAM, default 88MHz clock speed).


Defragster:
Did some testing with this - variation of above psram test sketch and also went to LittleFS in PSRAM and ran Integrity test.

Saw same faster 8MB perf and also faster on 16MB T_4.1 to verified tests passed - the LittleFS Integrity seemed 5% faster and no errors.

[Results noted here](https://forum.pjrc.com/threads/72572-Teensyduino-1-59-Beta-2?p=328746&viewfull=1#post328746) - that post #114 and #115

h4yn0nnym0u5e:
@A-Dunstan - if you want to bring this into sync with Paul's latest `cores`, I'll try to find time to make it conditional on whether the ISSI 16MByte PSRAM is in use. I've found a way to break the PSRAM test if that part and pre-fetch are both in use, which is probably not going to be popular!

A-Dunstan:
There are likely options somewhere in the FlexSPI config to prevent prefetching across a page boundary, which would be better.

h4yn0nnym0u5e:
I've had a trawl through the Reference Manual and internet, and I _think_ the i.MX RT1062 doesn't implement the read boundary capability. [This page](https://community.nxp.com/t5/i-MX-RT-Crossover-MCUs/FLEXSPI-prefetch-alignment/m-p/1810847) suggests not every NXP part implements it - also that you need to be careful in your choice of PSRAM, even the AP6404.

Looking at section 27.7.3.5, the AHBCR has plenty of separate "reserved" areas; [this page](https://ftp.rtems.org/pub/rtems/people/chrisj/releases/6/rc/6.1-rc8/docs/html/doxygen/group__FLEXSPI__Register__Masks.html) implies that it's bits 21 and 20 that would control the expected wrap size and treatment. But we don't have them. I can't even find which parts might have them ... but I guess that's academic.

A-Dunstan:
The IMXRT1062 manual says this:
``When AHBCR[READADDROPT] and AHBCR[PREFETCHEN] are both set to 1,
FlexSPI will guarantee flash access start address and data size are 64 bit aligned by
hardware.``
So that can be used to ensure the read address begins on a boundary that is 8-byte aligned. Additionally for the CPU each read address should already be 32-byte aligned due to the memory being cached.
_In theory_ I think this should be enough to configure AHBRXBUF0CR0 with the CPU master ID and BUFSZ of 4 (4*8 = 32), AHBRXBUFCR1 also with CPU master ID with BUFSZ set to use all the remaining buffers except 1 (I forget how many total there are, it's buried somewhere in the reference manual), AHBRXBUF2CR0 disabled and AHBRXBUF3CR0 enabled.
The idea is for AHBRXBUF1CR0 to tie up the unused buffer space (since all unallocated buffers go to AHBRXBUF3CR0), and it will never be used since AHBRXBUF0CR0 will match first for the CPU master ID.

The performance gain would probably be a lot less than before but it should still be better than nothing.

h4yn0nnym0u5e:
Hmm ... I'm not going to be able to give this the consideration it deserves for a couple of weeks, unfortunately. 

I can't find how big the AHB RX buffers are, either - it says it's in the "chip-specific section", but fails to provide a link... 

I haven't looked at them, but it's possible some magic is also needed for DMA access to PSRAM (which does seem to work for the AP6404 parts). I guess using BUF0 BUFSZ=4 for the core, and BUF1 BUFSZ=4 for eDMA, then tying up remaining-1 of the buffer in BUF2 might work. Paul used 64 for BUF0 and BUF1, but BUF2 and BUF3 appear to be set to illegal values (invalid MSTRID and possibly BUFSZ). 

It doesn't look as if there's any control over the TX buffer, even though page wrapping is a thing for Flash and PSRAM.

h4yn0nnym0u5e:
Candidate change PR now at https://github.com/A-Dunstan/cores/pull/1. It drops the speed improvement from 15% to 7%, but works with the ISSI chip as far as I can tell. It could be modified to allow the full speed improvement if the ISSI part isn't in use.

PR #713, opened 2023-08-08
Update some availableForWrite() declarations to be virtual

ssilverman:
These are overriding functions.

Note: Because this changes some core stuff, don't trust just me that it works. This should probably have at least one or two other people run tests.

### DIFF
--- a/teensy3/usb_seremu.h
+++ b/teensy3/usb_seremu.h
@@ -130,7 +130,7 @@ public:
 	size_t write(long n) { return 1; }
 	size_t write(unsigned int n) { return 1; }
 	size_t write(int n) { return 1; }
-	int availableForWrite() { return 0; }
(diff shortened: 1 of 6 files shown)


ssilverman:
By not declaring these as "overriding", if any objects are cast to a `Print`, then the wrong `availableForWrite()` function would get called.

ssilverman:
But this might be good to have, at least, for consistency.

ssilverman:
Spec details:

> Then this function in the class Derived is also virtual (whether or not the keyword virtual is used in its declaration) and overrides Base::vf (whether or not the word override is used in its declaration).

Source: https://en.cppreference.com/w/cpp/language/virtual

PaulStoffregen:
Looks like Arduino is now using virtual for availableForWrite(), so seem pretty safe to give this a try.

PR #714, opened 2023-08-10
add Print::vprintf() to teensy and teensy3

rnpnr:
This probably could have been included in 5021ecb but oh well.

Requested here: https://github.com/PaulStoffregen/cores/commit/5021ecb78828e81142864cef92788ad72dcc6494#commitcomment-124163913

### DIFF
--- a/teensy/Print.h
+++ b/teensy/Print.h
@@ -101,6 +101,7 @@ class Print
 	void clearWriteError() { setWriteError(0); }
 	int printf(const char *format, ...);
 	int printf(const __FlashStringHelper *format, ...);
+	int vprintf(const char *format, va_list ap) { return vdprintf((int)this, format, ap); }
   protected:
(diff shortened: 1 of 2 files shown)


ssilverman:
IMHO, I think this should just apply to Teensy 3 and not Teensy 2. The reasoning is that the Teensy 2 stuff hasn't been updated in a while and I think it was desired to not change anything. I'll leave this decision to Paul, however...

PaulStoffregen:
Will merge now and if Teensy 2 breaks, will undo later...

PaulStoffregen:
Indeed Teensy 2 needed a different approach.

https://github.com/PaulStoffregen/cores/commit/05e38fc903bc96a0e4b8a2620f3679f6d42a8ace

Just copying the Teensy 3 & 4 code caused all Teensy 2 programs to no longer compile.

Issue #561, opened 2021-05-28
`String::concat()` incompatible with Arduino's version


bblanchon:
Hi,

Arduino supports the following overloads of [`String::concat()`](https://www.arduino.cc/reference/en/language/variables/data-types/string/functions/concat/):

```c++
class String {
public:
  unsigned char concat(char c);
  unsigned char concat(const __FlashStringHelper *str);
  unsigned char concat(const char *cstr);
  unsigned char concat(const String &str);
  unsigned char concat(double num);
  unsigned char concat(float num);
  unsigned char concat(int num);
  unsigned char concat(long num);
  unsigned char concat(unsigned char c);
  unsigned char concat(unsigned int num);
  unsigned char concat(unsigned long num);

protected:
  unsigned char concat(const char *cstr, unsigned int length);
};
```

whereas Teensy defines:

```c++
class String {
public:
  String &concat(char c);
  String &concat(const __FlashStringHelper *pgmstr);
  String &concat(const char *cstr);
  String &concat(const String &str);
  String &concat(double num);
  String &concat(float num);
  String &concat(int num);
  String &concat(long num);
  String &concat(unsigned char c);
  String &concat(unsigned int num);
  String &concat(unsigned long num);
};
```

As you can see, Arduino returns a byte that indicates whether the operation succeeded or not.

It also defines an overload that accepts the size of the string as a second parameter.
I know this overload is `protected` in most cores (it's `public` in others), but if you want to be fully compatible, you must also support derived classes.

I hope you can align Teensy's `String` interface with Arduino's one; it would really help writing portable code.

Best regards,
Benoit

bblanchon:
@RonPTS, what's up with the downvote?
Do you think that compliance with Arduino is a bad thing?

PR #715, opened 2023-08-12
Fix IPAddress cast and comparison operators to be const

ssilverman:
This will make it much easier to use these operators, especially when working with `const` `IPAddress`es, without having to `const_cast` away the `const` before using them. Also the Arduino version has these.

### DIFF
--- a/teensy3/IPAddress.h
+++ b/teensy3/IPAddress.h
@@ -73,16 +73,16 @@ public:
 
 	// Overloaded cast operator to allow IPAddress objects to be used where a pointer
 	// to a four-byte uint8_t array is expected
-	operator uint32_t () {
+	operator uint32_t () const {
 		return _address.dword;
 	}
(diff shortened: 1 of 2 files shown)


ssilverman:
See here for the Arduino AVR version:
https://github.com/arduino/ArduinoCore-avr/blob/master/cores/arduino/IPAddress.h

Note that the cast and comparison operators are `const` there.

I would also like to point out that the Teensy version defines `operator!=()`, while the Arduino one doesn't have this, so Teensy's `IPAddress` technically has different behaviour. Strict compatibility was used as a reason for not improving Teensy's version in the past, so should we remove Teensy's `operator!=()`?

ssilverman:
I'll also point out this other Arduino incompatibility (`String` class): https://github.com/PaulStoffregen/cores/issues/561

ssilverman:
Thanks for merging this. It will allow me to remove my need to define operator friend functions. :)

PR #722, opened 2023-09-30
Increase initial size of usb_string_serial_number by 2 on Teensy3/Teensy4

hmaarrfk:
I was auditing the code and noticed two things:

1. The initial size is 12, which is: not equal to 10 * 2, 0, or 10
2. The data is copied from a buffer of size 11, so the size should be at least 11.

I am unsure if the default value should be `bLength=24` or `bLength=2`, both seem somewhat valid to me.

### DIFF
--- a/teensy3/usb_desc.c
+++ b/teensy3/usb_desc.c
@@ -1678,9 +1678,9 @@ struct usb_string_descriptor_struct usb_string_product_name_default = {
         PRODUCT_NAME
 };
 struct usb_string_descriptor_struct usb_string_serial_number_default = {
-        12,
+        24,
         3,
-        {0,0,0,0,0,0,0,0,0,0}
+        {0,0,0,0,0,0,0,0,0,0,0,0}  // uint16_t
 };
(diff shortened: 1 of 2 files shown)


PaulStoffregen:
This is incorrect.  Please read the USB 2.0 spec regarding string descriptors.

hmaarrfk:
Will reopen. thank you.

hmaarrfk:
I see, i read that it is in fact 2 + "length of string".  The string stored is not "null terminated".

I still think that the value of "12" is off. it should be either "2" or "22" as I understand it.

hmaarrfk:
ultimately, it is inconsequential due to the initialization routine. thanks for the usb pointer.

PaulStoffregen:
> I still think that the value of "12" is off. it should be either "2" or "22" as I understand it.

screenshot

PaulStoffregen:
Oh, yes, now I see.  It should be 22.  If you still care about this, please send a new PR which just changes 12 to 22.

PR #731, opened 2024-01-18
don't set do_zero flag when initializing extmem heap (improve performance when allocating)

A-Dunstan:
Currently when the extmem_smalloc_pool is created, the do_zero flag is set which causes smalloc to always zero out allocated memory. This reduces allocation performance since it's only required by calls to calloc, not malloc.

This patch unsets the do_zero flag, manually clears any statically allocated EXTMEM memory (which wasn't being done before in violation of C spec - uninitialized static data should be zeroed) and makes extmem_calloc() call sm_**c**alloc_pool() instead of sm_**m**alloc_pool().


A-Dunstan:
Updated the patch; with do_zero==false smalloc only clears smalloc_hdr contents when memory is free'd. This makes allocations persist across reboots unless the pool is cleared at startup.

PR #739, opened 2024-04-14
Only define SG14_INPLACE_FUNCTION_THROW if not already defined

ssilverman:
In case another behaviour is needed.


ssilverman:
Another option is to rename this define to `TEENSY_INPLACE_FUNCTION_THROW` or `TEENSYDUINO_INPLACE_FUNCTION_THROW`.

ssilverman:
It's possible there's a different way. I'll experiment with that and close this for now.

PR #743, opened 2024-05-21
add missing overrides for global operator new

A-Dunstan:
Updating the toolchain to support C++17 added some new overloads for operators new and delete to support overly-aligned allocations. Overrides for these are missing from Teensyduino which results in the newlib defaults being used (which link in a bunch of exception throwing/handling stuff that we don't want).

(The delete overrides aren't all that important, they're just there to ensure any memory allocated with our new operators is deallocated using the appropriate method i.e. via free() ).

### DIFF
--- a/teensy4/new.cpp
+++ b/teensy4/new.cpp
@@ -23,7 +23,8 @@
  * SOFTWARE.
  */
 
-#include <stdlib.h>
+#include <cstdlib>
+#include <new>
 
 void * operator new(size_t size)
 {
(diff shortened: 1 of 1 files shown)


PaulStoffregen:
I'm concerned about changing the includes.  Is that really necessary?

A-Dunstan:
For the stdlib inclusion, .cpp files should use the c++ include files.
Iirc adding `#include <new>` was necessary to get the std::align_val_t definition.

PR #745, opened 2024-06-05
Add printf format compiler checks for Print::printf()

ssilverman:
This will help catch unintended printf() arguments.

### DIFF
--- a/teensy3/Print.h
+++ b/teensy3/Print.h
@@ -111,9 +111,7 @@ class Print
 	size_t println(const Printable &obj)		{ return obj.printTo(*this) + println(); }
 	int getWriteError() { return write_error; }
 	void clearWriteError() { setWriteError(0); }
-	// format warnings are too pedantic - disable until newer toolchain offers better...
-	// https://forum.pjrc.com/threads/62473?p=256873&viewfull=1#post256873
-	int printf(const char *format, ...) /*__attribute__ ((format (printf, 2, 3)))*/;
+	int printf(const char *format, ...) __attribute__((format(printf, 2, 3)));
 	int printf(const __FlashStringHelper *format, ...);
 	int vprintf(const char *format, va_list ap) { return vdprintf((int)this, format, ap); }
   protected:
--- a/teensy3/usb_seremu.h
+++ b/teensy3/usb_seremu.h
@@ -36,6 +36,7 @@
 #if defined(SEREMU_INTERFACE)
 
 #include <inttypes.h>
+#include <stdarg.h>
 
 #if F_CPU >= 20000000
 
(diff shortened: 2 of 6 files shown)


ssilverman:
Here is an example of a program that should, but doesn't, signal compiler warnings for `Serial.printf()` calls:

```c++
#include <cstdio>

class X {};

void setup() {
  Serial.begin(115200);
  while (!Serial && millis() < 4000) {
    // Wait for Serial
  }

  X x;

  // The following do not build with a warning; they should
  Serial.printf("%u\r\n", Serial);
  Serial.printf("%u\r\n", x);

  // The following do build with a warning
  std::printf("%u\r\n", Serial);
  std::printf("%u\r\n", x);
}

void loop() {
}
```

The former code calls out this forum post: https://forum.pjrc.com/threads/62473?p=256873&viewfull=1#post256873

I fundamentally disagree with the reasons against using these warnings in that post. The types `unsigned int` and `unsigned long` _most definitely are not_ the same type, even if they both happen to be the same size. Any warnings produced by the compiler are generally legitimate. If they're considered too pedantic or verbose, then that's just because of a more "relaxed" coding style. C++ itself is very strongly typed, and so proper use of `printf()` and any compiler warnings or errors should reflect that.

If these warnings had been there, problems such as those mentioned in the following forum post would have been found much quicker: https://forum.pjrc.com/index.php?threads/id-like-to-file-a-complaint-s.75146/

PaulStoffregen:
We tried this years ago.  It was far too pedantic about 32 bit long versus 32 bit int.

PaulStoffregen:
I will only merge this if we can find a way for Serial.printf("%d", (long)number) and Serial.printf("%ld", (int)number) to avoid needless warnings.

ssilverman:
Those two examples _should_ print a warning, because `%d` is not meant for the `long` type, and `%ld` is not meant for the `int` type.

May I ask: why would you not want these to print warnings? Is it to maintain the "Arduino ease of use" thing?

ssilverman:
I think I know how we might be able to prevent `Serial`, specifically, to not print those warnings, but it will add a little code. I'll add a commit shortly...

PaulStoffregen:
Let's continue this on the forum thread.  I'll reply there.

ssilverman:
I added a second commit showing how `Serial` could be made to not print `printf` format errors. In addition, I'll make this PR a draft for now.

Unfortunately, there's lots of boilerplate code, and you may not want these additions.

Some of my reasons for wanting `Print::printf()` to catch format warnings is that otherwise, it obligates every single subclass to not make use of warnings. I would have saved myself lots of debugging time if I was able to have all my `Print` subclasses not be forced to not show warnings. Instead, I think what's really wanted is the most common use of `Serial` to not cause warnings for mismatched format strings. (I think that's unfortunate, but it's what seems to be wanted here.)

Sure, we can continue on the forum...

ssilverman:
Note also, that for the second commit, it may be desirable to add for `usb_serial2_class` and `usb_serial3_class`, as well as the USB serial class definitions under the "usb_*" directories.

PR #752, opened 2024-08-28
remove XBAR1_IN10 from pin_to_xbar_info

sndsgd:
Searching the reference manual for `XBAR1_IN10` yields a single line, but there are no occurences of `XBAR1_IN10_SELECT_INPUT`.
2024-08-28-130023-screenshot

Additionally, it appears that `pin_to_xbar_info` is only used in `HardwareSerial` and there is no uart associated with `GPIO_B0_12`:
2024-08-28-130244-screenshot


A-Dunstan:
This doesn't make sense to me.
XBAR1_INOUT10 is Teensy pin 32. It has no alternative pins, so it has no SELECT_INPUT register. It's still usable as an XBAR pin which means it's usable by HardwareSerial.

sndsgd:
@A-Dunstan thanks for the response! A few things led to me cutting this PR:

1. A lack of understanding what XBAR is and how it can be used.
1. Upon testing using pin 32 as an alternate rx pin, I my application stopped rx-ing data, and eventually stalled. I should have tested more pins :sweat_smile: 

I had some time to go over this today, and it looks like there is a bug regarding `setRx()` and _all_ xbar pins. PR to address is here: https://github.com/PaulStoffregen/cores/pull/754

PR #755, opened 2024-09-09
AudioStream: Add destructor.  Enables dynamic creation and destruction.

chipaudette:
There's been great work on dynamic creation and destruction of AudioConnection instances.  But, one has not been able to dynamically destroy AudioStream instances.  In my view, the main limitation is that the user had no way of removing an AudioStream instance from the AudioStream update list.

This pull request adds a destructor to AudioStream.  Now, when the instance is destroyed, it removes the pointer from the update list.

The user must still remember to close out all of the AudioConnections that use the destroyed AudioStream instance, but that's at least possible for the user to do.  Without AudioStream having a destructor, there is no way for the user to remove the instance to the update list.

This pull request includes both Teensy3 and Teensy4.  It has been tested on a Teensy 4.1 and a Teensy 3.6.

### DIFF
--- a/teensy3/AudioStream.cpp
+++ b/teensy3/AudioStream.cpp
@@ -208,6 +208,31 @@ audio_block_t * AudioStream::receiveWritable(unsigned int index)
 	return in;
 }
 
+//Define the destructor for AudioStream.  
+//The contract here is that the user has already disconnected all or
+//destroyed all of the AudioConnections.  Therefore, the clean-up activities
+//that remain are to remove this instance from the update list and
+//to release any audio_block_t pointers in this instance's input queue.
+AudioStream::~AudioStream(void)
+{
+	//remove this instance from the update list
+	__disable_irq(); //make sure no update() happens while we're changing the update list
+	AudioStream *p = first_update;
+	while (p != NULL) {
+		if (p->next_update == this) {
+			p->next_update = this->next_update; //remove the pointer to this instance and replace it with the next instance
+		}
+		p = p->next_update; //go to next step in linked list
+	}
+	__enable_irq();
+
+	//release any audio blocks held in the input queue for this instance of AudioStream
+	for (int i=0; i<num_inputs; i++) {
+		audio_block_t *block = inputQueue[i];
+		if  (block != NULL) release(block);
+	}
+}
+
 /**************************************************************************************/
 // Constructor with no parameters: leave unconnected
 AudioConnection::AudioConnection()
--- a/teensy4/AudioStream.cpp
+++ b/teensy4/AudioStream.cpp
@@ -200,6 +200,31 @@ audio_block_t * AudioStream::receiveWritable(unsigned int index)
 	return in;
 }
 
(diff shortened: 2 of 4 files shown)


teensy3/AudioStream.cpp
```diff
@@ -208,6 +208,31 @@ audio_block_t * AudioStream::receiveWritable(unsigned int index)
...
+		if (p->next_update == this) {
+			p->next_update = this->next_update; //remove the pointer to this instance and replace it with the next instance
+		}
+		p = p->next_update; //go to next step in linked list
+	}
+	__enable_irq();
+
+	//release any audio blocks held in the input queue for this instance of AudioStream
+	for (int i=0; i<num_inputs; i++) {
+		audio_block_t *block = inputQueue[i];
```
h4yn0nnym0u5e, line 231: This doesn't work - the inputQueue[] array is part of the _derived_ class, and has _already been destroyed_ by the time this code executes

teensy4/AudioStream.cpp
```diff
@@ -200,6 +200,31 @@ audio_block_t * AudioStream::receiveWritable(unsigned int index)
...
+		if (p->next_update == this) {
+			p->next_update = this->next_update; //remove the pointer to this instance and replace it with the next instance
+		}
+		p = p->next_update; //go to next step in linked list
+	}
+	__enable_irq();
+
+	//release any audio blocks held in the input queue for this instance of AudioStream
+	for (int i=0; i<num_inputs; i++) {
+		audio_block_t *block = inputQueue[i];
```
h4yn0nnym0u5e, line 223: This doesn't work - the inputQueue[] array is part of the _derived_ class, and has _already been destroyed_ by the time this code executes

A-Dunstan:
I think the list removal has a couple of errors:
- it only ever checks if next_update == this, so the stream pointed to by first_update can never be unlinked
- since any stream instance will only be in the list once, there should be a break statement to terminate the loop once it's found/unlinked

teensy4/AudioStream.cpp
chipaudette: Thanks for pointing this out.  It's been removed from this minimal pull request and replaced with comment.   The likely slow leak of AudioMemory should be addressed through a subsequent pull request.

chipaudette:
I believe that I have now addressed the last two bullets ("first_update" and "break").  All of the latest edits have been tested on Teensy 4.1 and Teensy 3.6.

chipaudette:
I'm plugging for this Pull Request again.  Presuming that one trusts Bjarne Stroustrup and Herb Sutter, their ISO C++ guidelines, a base class destructor should be public and virtual (or protected and non-virtual). 

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c35-a-base-class-destructor-should-be-either-public-and-virtual-or-protected-and-non-virtual

This pull request is giving AudioStream a destructor that is public and virtual in line with this ISO C++ guidance.

Any chance it could be accepted?

PR #756, opened 2024-09-21
Stop hardcoding WinUSB descriptor string lengths

forderud:
Done to reduce risk of the descriptor lengths getting out of sync.


forderud:
@PaulStoffregen The background for "bombarding" you with PR's is that I'm planning to modify the USB implementation in this repo for a custom HW device feasibility project on Teensy. As part of that I'm interested in upstreaming some generic SW improvements to your USB sources, so that they can also benefit others. It will also prevent my fork from diverging more than necessary.

I will of course test the changes locally, split the changes into small and easy to review PR's (like done here) and explain the rationale for each change. I will furthermore do my best to accommodate review feedback in due time.

Are you interested in receiving more PR's like this?

PR #761, opened 2024-09-22
[USB] Document composite device with IAD

forderud:
Document that `DEVICE_CLASS=0xEF`, `DEVICE_SUBCLASS=0x02` & `DEVICE_PROTOCOL=0x01` is used to encode a composite device with IAD, since this is not obvious from reading the code.

Documentation: https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/usb-interface-association-descriptor


forderud:
Maybe if could be an idea to also document that `DEVICE_CLASS=0x00`, `DEVICE_SUBCLASS=0x00` & `DEVICE_PROTOCOL=0x00` encodes a composite device _without_ IAD(?)

PR #765, opened 2024-09-27
HardwareSerial::attachRts - new optional parameter to invert signal

KurtE:
Added  parameter to attachRts that if set will invert the signals:
```
bool attachRts(uint8_t pin, bool invert_signal=false);
```

This should allow more flexibility in connecting two teensy 4.x boards to each other and using the CTS/RTS between the two boards for flow control.

This is in response to the thread:
https://forum.pjrc.com/index.php?threads/teensy-4-0-4-1-serial1-attachcts.75899/#post-350107

Where I felt somewhat strange to require someone to have to use hardware inverters to do this.

### DIFF
--- a/teensy4/HardwareSerial.cpp
+++ b/teensy4/HardwareSerial.cpp
@@ -244,11 +244,13 @@ void HardwareSerialIMXRT::begin(uint32_t baud, uint16_t format)
 
 inline void HardwareSerialIMXRT::rts_assert()
 {
-	DIRECT_WRITE_LOW(rts_pin_baseReg_, rts_pin_bitmask_);
+	if (rts_pin_invert_) DIRECT_WRITE_HIGH(rts_pin_baseReg_, rts_pin_bitmask_);
+	else DIRECT_WRITE_LOW(rts_pin_baseReg_, rts_pin_bitmask_);
 }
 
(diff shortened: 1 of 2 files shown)


PaulStoffregen:
I'm hoping we can avoid more true / false parameters in public APIs.  Main issue is they lead to less readable user code, where you would need to look up in the documentation what the true vs false input means to understand what the program is doing.

In this case, let's go with a descriptive function name like attachRtsInverted().

KurtE:
We can do... But

Personally, I believe that this:
>In this case, let's go with a descriptive function name like attachRtsInverted().

Is equally unreadable...  Inverted from What?

If we really wanted readability, we could do something like:

```
typedef enum {
    RTS_ASSERTED_PIN_SIGNAL_LOW = 0,  // default
    RTS_ASSERTED_PIN_SIGNAL_HIGH = 1
} RTS_assert_pin_level_t;

bool attachRts(uint8_t pin, RTS_assert_pin_level_t assert=RTS_ASSERTED_PIN_SIGNAL_LOW); 
```
Warning typed on fly so probably typos.  And the comments above the definition should then
change to match.

But your call

KurtE:
I pushed up changes to match the API name you requested...

PR #771, opened 2024-11-29
Change 'virtual' to 'override' where appropriate

ssilverman:
This makes it clearer which functions override.


ssilverman:
I'll un-draft this as soon as #772 gets in.

PR #772, opened 2024-11-29
Fix File to mark overridden functions as 'override'

ssilverman:
Enables a File object to be used via a pointer or reference to Stream
and Print.


ssilverman:
Note to reviewer: I can use 'virtual' instead, if you like, but this PR will change all overridden methods to use 'override' instead of 'virtual': #771.

A-Dunstan:
>  Without 'virtual' or 'override', the default Stream and Print functions will be used instead.

This shouldn't be the case. If a derived class implements a virtual method from the base class it automatically replaces it without any extra notation required. Specifying "override" is just a precaution that will make the compiler throw an error if there isn't a matching virtual base method.

ssilverman:
> > Without 'virtual' or 'override', the default Stream and Print functions will be used instead.
> 
> This shouldn't be the case. If a derived class implements a virtual method from the base class it automatically replaces it without any extra notation required. Specifying "override" is just a precaution that will make the compiler throw an error if there isn't a matching virtual base method.

Ooops, you're right. I'll update the PR description and title. Overridden/virtual/etc functions should still be marked as such. It's good practice so that people reading the code don't have to guess or research every function.

Relevant portions of the spec:
https://en.cppreference.com/w/cpp/language/virtual:
> Then this function in the class Derived is also virtual (**whether or not the keyword virtual is used in its declaration**) and overrides Base::vf (whether or not the word override is used in its declaration).

PR #778, opened 2025-03-20
Fix alias definitions for FLEXIO_TIMCFG_DISABLE_* to match reference manual

brandondahler:
### Problem
Definitions for `FLEXIO_TIMCFG_DISABLE_ON_TRIGGER_RISING` and `FLEXIO_TIMCFG_DISABLE_ON_TRIGGER_CHANGE` are incorrectly named and invalid (respectively) per the [IMXRT1060RM Rev 3 reference manual](https://www.pjrc.com/teensy/IMXRT1060RM_rev3.pdf), section 50.6.1.21.4, page 3004.

#### Details
Per the manual, TIMDIS has these values:
```
14-12
TIMDIS
Timer Disable
Configures the condition that causes the Timer to be disabled and stop decrementing.
000b - Timer never disabled
001b - Timer disabled on Timer N-1 disable
010b - Timer disabled on Timer compare (upper 8-bits match and decrement)
011b - Timer disabled on Timer compare (upper 8-bits match and decrement) and Trigger Low
100b - Timer disabled on Pin rising or falling edge
101b - Timer disabled on Pin rising or falling edge provided Trigger is high
110b - Timer disabled on Trigger falling edge
111b - Reserved
```

* `FLEXIO_TIMCFG_DISABLE_ON_TRIGGER***_RISING***` is configured to represent `FLEXIO_TIMCFG_TIMDIS(6)`, which is `110b - Timer disabled on Trigger ***falling*** edge`
* `FLEXIO_TIMCFG_DISABLE_ON_TRIGGER_CHANGE` is configured to represent `FLEXIO_TIMCFG_TIMDIS(7)`, which is `111b - Reserved`

### Proposed Solution
* A new `#define FLEXIO_TIMCFG_DISABLE_ON_TRIGGER_FALLING  FLEXIO_TIMCFG_TIMDIS(6)` has been added to fix the naming
* The existing `FLEXIO_TIMCFG_DISABLE_ON_TRIGGER_RISING` and `FLEXIO_TIMCFG_DISABLE_ON_TRIGGER_CHANGE` definitions remain as-is, but a deprecation comment has been appended similar to `__CM7_CMSIS_VERSION_MAIN`'s definition
  * Figured there's no harm in leaving them as is so that we don't break anyone's code if they somehow happen to be using the existing definitions in spite of their incorrect naming.


A-Dunstan:
I don't see why DISABLE_ON_TRIGGER_CHANGE should be left (even deprecated) if the value is incorrect.

brandondahler:
Fair point, I was thinking about both deprecations, but for that specific one you're fully right -- it probably causes the timer to be broken if that value is actually used so a build failure is better than a silent deprecation.  Will push an update later to remove it fully.

PR #782, opened 2025-05-15
Add new PSRAM ID and sizes

xxxajk:
Adds ISSI PSRAM, with ability to use 4MB/8MB/16MB chips.
Limit is still 16MB (see comments) but should be able to support 64MB via two 32MB PSRAM.
Changed how the region is determined to make all decisions on sizes in one place.


h4yn0nnym0u5e:
As #784 has been merged, presumably this can be closed?

PR #790, opened 2025-11-04
Simplify ability to create custom USB devices including MTP

h4yn0nnym0u5e:
Choices made in [this commit](https://github.com/PaulStoffregen/cores/commit/bb9d9b9c15bdb22fb7de1218c0161998afed55c1) relate to specific choices on the Arduino IDE Tools menu, rather than whether the MTP interface has been included as a consequence of the choice. This is inconsistent with other interfaces, e.g. MIDI, keyboard, Serial (CDC) etc. As a result, a custom mix of interfaces required edits to be made to the MTP sources as well (see [this forum post](https://forum.pjrc.com/index.php?threads/how-to-teensy-as-all-in-one-serial-midi-keyboard-mouse-audio-mtp.77430/post-362830)).

This PR adopts the system used for other interfaces. It has been briefly tested, and appears to work OK.


h4yn0nnym0u5e:
Did a check over cores for other places where a specific Tools menu selection was checked, rather than the availability of a USB interface, and found a couple relating to MIDI and Serial.

h4yn0nnym0u5e:
Commit https://github.com/PaulStoffregen/cores/commit/4591d5c1ac9769f7e2b9ad77afcfd09f96002dee covers some of this, but not all of it...

PR #791, opened 2025-11-29
Add fgets to File Impl in FS.h

mjs513:
Ran across an issue that fgets was not supported.  Looking through the forum seems like a few other people ok maybe one or two.  But it also looks like a handy function to have available so added it.  You can look here for a quick test: https://forum.pjrc.com/index.php?threads/problems-with-external-library-using-sdfat.77492/post-363551 

Will make a similar change to the SD library.


A-Dunstan:
This is not something that should be implemented in the File/FS classes, the real (libc) fgets function is usable by properly implementing _read() and _write() to use File objects as int file descriptors (fds).

(This is also a bad implementation - doesn't check if f is valid before calling the fgets virtual method on it.)

ssilverman:
I’m with @A-Dunstan on this one. I don’t think this belongs in File either.

PR #792, opened 2025-12-09
Fix compilation error with recent newlib versions

sebastiendu:
because random() return value doesn't match.

This reverts commit ff83ba55

### DIFF
--- a/teensy3/WMath.cpp
+++ b/teensy3/WMath.cpp
@@ -33,9 +33,9 @@ void srandom(unsigned int newseed)
 	seed = newseed;
 }
 
-int32_t random(void)
+long random(void)
 {
-	int32_t hi, lo, x;
+	long hi, lo, x;
 
 	// the algorithm used in avr-libc 1.6.4
 	x = seed;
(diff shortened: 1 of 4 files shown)


A-Dunstan:
What is this meant to fix exactly? The topic is the same as the commit that it is reverting, which makes no sense.

sebastiendu:
It fixes the following error:

```
In file included from cores/teensy3/Arduino.h:6:
cores/teensy3/WProgram.h:88:9: error: functions that differ only in their return type cannot be overloaded
   88 | int32_t random(void);
      | ~~~~~~~ ^
/usr/include/newlib/stdlib.h:260:6: note: previous declaration is here
  260 | long    random (void);
      | ~~~~    ^
1 error generated.
```

A-Dunstan:
In the build of newlib that Teensy cores is meant to be used with, int32_t and long are equivalent types. Your "/usr/include/newlib" copy seems to be configured differently.

sebastiendu:
My newlib is the standard, Debian stable version.

But I agree this is not the standard Teensy development environment because I have been using clang-19 instead of the old gcc toolchain. I remember gcc did not complain about this. clang-19 is more strict than gcc. It knows long and int32_t are the same, but still wants the signature to be literally the same.

I believe this patch would not break anything, and fix a minor error in the core files, therefore possibly benefit others.

Not MANY others however. I might even be the only one using clang to compile a Teensyduino microcode with newlib.

So feel free to reject this pull request, I will just keep on using my own fork.

PR #793, opened 2026-01-13
Improve uses of 'static inline' and 'inline'

ssilverman:
1. When C++, use just `inline` instead of `static inline` for external
   functions, to ensure there's only a single copy of those functions.
   In other words, make these definitions have external linkage. This
   improvement also lets functions that are intended to have external
   linkage, such as `millis()`, be used as template arguments without
   complaints about anonymous namespaces or internal linkage.
2. Keep `static inline` for C-only functions and when there wasn't a
   check for `__cppreference` in the file.
3. Remove `inline` for functions declared or defined within a class
   declaration. Those are implicitly "inline".

The problem this solves is that it ensures there's only one linked
version of a function instead of multiple copies of that function.
This will save a little space.
    
Note that this change only affects the headers. It's possible for
a header to be included in more than one translation unit, meaning
that if a function has its definition in a header, there will be
duplicate copies of that function. Here's where `static` and `inline`
come in.
    
The `static` part means "internal linkage", where the function may be
included in several compiled object files and linked without conflict.
'inline' in modern C++ came to mean "multiple definitions are allowed"
(as long as they're the same), which can happen if functions are
defined in a header that's included in more than one translation unit.
If just `inline` without `static` is used, then a potentially-
conflicting multiply-defined function is okay and the linker will
choose one of them.
    
One example is the `millis()` function. It's declared `static inline`
and therefore will likely have multiple copies in the final binary.
Note that optimization settings may change this, but regardless, by
being declared `static`, the function has "internal linkage", so there
is effectively one copy per object file.

More information can be found here:
https://en.cppreference.com/w/cpp/language/inline.html

For the functions also accessed from C, it's still appropriate to keep
the `static inline`. More information can be found here:
https://en.cppreference.com/w/c/language/inline.html

### DIFF
--- a/teensy/HardwareSerial.h
+++ b/teensy/HardwareSerial.h
@@ -7,7 +7,7 @@
 class HardwareSerial : public Stream
 {
 public:
-	inline void begin(uint32_t baud, uint8_t txen_pin=255) {
+	void begin(uint32_t baud, uint8_t txen_pin=255) {
 		_begin(((F_CPU / 8) + (baud / 2)) / baud, txen_pin);
 	}
 	void _begin(uint16_t baud_count, uint8_t pin);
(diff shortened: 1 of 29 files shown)


ssilverman:
See also: https://en.wikipedia.org/wiki/Inline_(C_and_C%2B%2B)

PR #794, opened 2026-02-04
teensy4/startup: Put default time into a const variable

ssilverman:


A-Dunstan:
With the compiler recently updated, is there a way to parse `__DATE__` and `__TIME__` as a constexpr to use those as the default? I have a vague memory that newer c++ was needed to handle strings at compile time.

ssilverman:
> With the compiler recently updated, is there a way to parse `__DATE__` and `__TIME__` as a constexpr to use those as the default? I have a vague memory that newer c++ was needed to handle strings at compile time.

I think this is a good idea. The only potential issue I could see with this is if this file doesn't get rebuilt every time the firmware is built.

PaulStoffregen:
On Teensy 3.6 in boards.txt we have this for the linker command:

```
teensy36.build.flags.ld=-Wl,--gc-sections,--relax,--no-warn-rwx-segments,--defsym=__rtc_localtime={extra.time.local} "-T{build.core.path}/mk66fx1m0.ld"
```

Arduino's build system substitute unix time for "{extra.time.local}".  It's actually used to initialize the RTC.

https://github.com/PaulStoffregen/cores/blob/master/teensy3/mk20dx128.c#L1136

On Teensy 4.x this approach was abandoned in favor of having Teensy Loader send the time and the bootloader init the RTC, so there's much less latency and the RTC doesn't get initialized again (incorrectly) if Teensy reboots and the RTC had stopped for any reason.

We could probably add this linker defined symbol on Teensy 4 also, even though it wouldn't be used for RTC init.

PR #797, opened 2026-02-23
Update feedback_accumulator calculation for audio to use AUDIO_SAMPLE_RATE_EXACT

bthayer3:
I found this formula was mentioning 44.1 kHz which could be improved by referencing the exact variable. With this change, `feedback_accumulator` value is more dynamic and respects changes to AUDIO_SAMPLE_RATE_EXACT.

Note I only am updating Teensy 4 version. I don't know the formula used for Teensy 3 version; I am not sure of the formula for this and how 185042824 was determined.


djphazer:
FYI, a similar change is lined up in this branch along with a bunch of other fixes for multi-channel USB Audio: https://github.com/h4yn0nnym0u5e/cores/tree/alex6679/add-multi-channel-usb

